PR #1, opened 2019-10-04
Add Creative Commons 0 license

BatmanAoD:
I suggest CC0 because (1) I don't expect this repo to have code, and (2) I don't expect sharing without attribution to be a problem

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,116 @@
+CC0 1.0 Universal
+
+Statement of Purpose
+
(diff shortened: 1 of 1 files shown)


nikomatsakis:
I would suggest we use the same license as the [rust-rfcs repository](https://github.com/rust-lang/rfcs/) -- we tend to just use the same dual-licensing scheme there for overall simplicity.

BatmanAoD:
Will submit separate PRs for Apache & MIT licenses

PR #2, opened 2019-10-04
Fill in some text

BatmanAoD:
The new content is mostly based on the [HackMD doc](https://hackmd.io/arj3QG1dRPOOM1T5BIuv_w).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,16 @@
 # project-ffi-unwind
 
-A project to extend the Rust language to support unwinding that crosses FFI boundaries.
+A [working-group project][shepherds-blog] to extend the Rust language to
+support unwinding that crosses FFI boundaries.
 
-- [What is a project?](http://smallcultfollowing.com/babysteps/blog/2019/09/11/aic-shepherds-3-0/)
 - Shepherds:
   - [acfoltzer(Alex)](https://github.com/acfoltzer)
   - [batmanaod (Kyle)](https://github.com/batmanaod)
-  - with guidance/assistance from [nikmoatsakis (Niko)](https://github.com/nikmoatsakis)
+- Rust lang team contacts:
+  - [nikmoatsakis (Niko)](https://github.com/nikmoatsakis)
+  - [joshtriplett (Josh)](https://github.com/joshtriplett)
 - [Our charter](charter.md)
 - [Our plan](plan.md)
 - [Cross-language unwinding FAQ](faq.md)
+
+[shepherds-blog]: http://smallcultfollowing.com/babysteps/blog/2019/09/11/aic-shepherds-3-0/
--- a/README.md
+++ b/README.md
@@ -4,9 +4,9 @@ A project to extend the Rust language to support unwinding that crosses FFI boun
 
 - [What is a project?](http://smallcultfollowing.com/babysteps/blog/2019/09/11/aic-shepherds-3-0/)
 - Shepherds:
-  - [acfoltzer](https://github.com/acfoltzer)
+  - [acfoltzer(Alex)](https://github.com/acfoltzer)
   - [batmanaod (Kyle)](https://github.com/batmanaod)
-  - with guidance/assistance from [nikmoatsakis](https://github.com/nikmoatsakis)
+  - with guidance/assistance from [nikmoatsakis (Niko)](https://github.com/nikmoatsakis)
 - [Our charter](charter.md)
 - [Our plan](plan.md)
 - [Cross-language unwinding FAQ](faq.md)
--- a/README.md
+++ b/README.md
@@ -9,8 +9,10 @@ support unwinding that crosses FFI boundaries.
 - Rust lang team contacts:
   - [nikmoatsakis (Niko)](https://github.com/nikmoatsakis)
   - [joshtriplett (Josh)](https://github.com/joshtriplett)
+- [Our chat room][zulip-room]
 - [Our charter](charter.md)
 - [Our plan](plan.md)
 - [Cross-language unwinding FAQ](faq.md)
 
 [shepherds-blog]: http://smallcultfollowing.com/babysteps/blog/2019/09/11/aic-shepherds-3-0/
+[zulip-room]: https://rust-lang.zulipchat.com/#narrow/stream/210922-wg-ffi-unwind/topic/welcome/near/177543226
--- a/README.md
+++ b/README.md
@@ -1,1 +1,12 @@
 # project-ffi-unwind
+
+A project to extend the Rust language to support unwinding that crosses FFI boundaries.
+
+- [What is a project?](http://smallcultfollowing.com/babysteps/blog/2019/09/11/aic-shepherds-3-0/)
+- Shepherds:
+  - [acfoltzer](https://github.com/acfoltzer)
+  - [batmanaod (Kyle)](https://github.com/batmanaod)
+  - with guidance/assistance from [nikmoatsakis](https://github.com/nikmoatsakis)
+- [Our charter](charter.md)
+- [Our plan](plan.md)
+- [Cross-language unwinding FAQ](faq.md)
--- a/faq.md
+++ b/faq.md
@@ -1,0 +1,32 @@
+## FAQ questions
+
+- Why use a new ABI? Why not an attribute?
+  - Unlike ABIs, there is no syntax for assigning attributes to function
+    pointers. The ABI must be part of the type system so that callers of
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -1 +1,12 @@
...
\ No newline at end of file
+# project-ffi-unwind
+
+A project to extend the Rust language to support unwinding that crosses FFI boundaries.
+
+- [What is a project?](http://smallcultfollowing.com/babysteps/blog/2019/09/11/aic-shepherds-3-0/)
+- Shepherds:
+  - [acfoltzer](https://github.com/acfoltzer)
+  - [batmanaod (Kyle)](https://github.com/batmanaod)
+  - with guidance/assistance from [nikmoatsakis](https://github.com/nikmoatsakis)
```
BatmanAoD:
```suggestion
  - with guidance/assistance from [Niko](https://github.com/nikmoatsakis)
```

I think having our "handles" and our names is probably a good idea, though I think that matters most in my case, where there's no clear connection between my name and my handle.

faq.md
```diff
@@ -0,0 +1,32 @@
+## FAQ questions
+
+- Why use a new ABI? Why not an attribute?
```
nikomatsakis: Formatting nit: I think let's use `###` headings (this also means people can link to the questions)

nikomatsakis (changes requested):

nikomatsakis (approved):

faq.md
```diff
@@ -0,0 +1,31 @@
...
+- mark declarations in Rust `extern "C unwind"`
+- Do *not* use `catch_unwind` in calling code
+- compile w/ `panic = unwind`
+- Check compiler, linker, & platform documentation (TODO: expand on this)
+- Re-test for every compiler update
+- (TODO: other caveats?)
+### How does cross-language unwinding differ from cross-language
+  `setjmp`/`longjmp`?
+- `setjmp`/`longjmp` across Rust frames is currently guaranteed not to have
+  undefined behavior or to `abort`
```
gnzlbg, line 17: This is a weird way to say that the behavior of `setjmp`/`longjmp` is defined. Maybe just explain how their behavior is defined instead?

faq.md
```diff
@@ -0,0 +1,31 @@
...
+- It should never be assumed that `drop` will be called for objects in
+  intermediate frames traversed by a `longjmp`, but this may occur on certain
+  platforms. Rust provides no guarantee either way. Conversely, on platforms
+  where unwinding is supported, it will always call `drop`.
```
gnzlbg, line 21: What does this last sentence mean ? E.g. `x86_64-unknown-linux-gnu` is a platform where unwinding is supported, but longjmp never drops values living in stack frames. On other platforms like MSVC, do `longjmp`s always drop values on stack frames?

faq.md
```diff
@@ -0,0 +1,31 @@
...
+- compile w/ `panic = unwind`
+- Check compiler, linker, & platform documentation (TODO: expand on this)
+- Re-test for every compiler update
+- (TODO: other caveats?)
+### How does cross-language unwinding differ from cross-language
+  `setjmp`/`longjmp`?
+- `setjmp`/`longjmp` across Rust frames is currently guaranteed not to have
+  undefined behavior or to `abort`
+- It should never be assumed that `drop` will be called for objects in
+  intermediate frames traversed by a `longjmp`, but this may occur on certain
```
gnzlbg, line 19: If replacing a `longjmp` with a `panic` and a `setjmp` with a `catch_unwind` causes any destructors to run, the behavior is undefined. Some safe Rust abstractions like `Pin` rely on such things never happening.

faq.md
```diff
@@ -0,0 +1,31 @@
...
+  platforms. Rust provides no guarantee either way. Conversely, on platforms
+  where unwinding is supported, it will always call `drop`.
+- Rust does not have a concept of `Copy` for stack-frames, which would permit
+  the compiler to check that `longjmp` may safely traverse those frames. Such a
+  language feature may be added in the future, but although it would be useful
```
gnzlbg, line 24: You can point to https://github.com/Centril/rfc-effects/issues/11 if you want to add more context about what that would entail.

faq.md
```diff
@@ -0,0 +1,31 @@
...
+  for `longjmp`, it would not be useful for unwinding.
+- `setjmp`/`longjmp` does not involve data transfer (the exception object)
```
gnzlbg, line 26: The return value of `setjmp` is a data-transfer from a `longjmp`.

faq.md
```diff
@@ -0,0 +1,31 @@
...
+- unwinding involves the use of a personality function, which raises additional
+  cross-language compatibility concerns; `setjmp`/`longjmp` does not
```
gnzlbg, line 28: The last sentence is unfinished.

gnzlbg (commented):
The setjmp/longjmp section looks good but there are some minor issues open. 

Also, I'm not sure that the claim that drop is always guaranteed to run in some platforms is true. I recall that on MSVC whether objects are drop on a longjmp depends on whether a function is inlined or not. For example, if you have something like:

```rust
fn foo() {
   if setjmp(jmpbuf) == 0 {
       bar()     
   }
}
fn bar() {
   let x: DropT;
   baz()
}
fn baz() { longjmp(jmpbuf, 1) }
```

and `bar` is not inlined, then the destructor of `bar::x` will run because the bar function is unwound by the longjmp implementation. However, if `bar` is inlined into `foo`, the `longjmp` does not unwind over the cleanup code of `bar::x` because it will be somewhere in `foo`s stack frame, and that stack frame is not unwound, so the destructor for `bar::x` does not run.

faq.md
BatmanAoD: "Not undefined" is not equivalent to "is defined"; this is an (arguably unfortunate) legacy of the C/C++ ISO standards. Nico and I agree that Rust needs to better specify how "unspecified", "undefined", "implementation defined", etc are to be used in the Rust Reference and elsewhere in the Rust community, but at the moment, I felt that this was the safest phrasing, because "undefined behavior" is (ironically) the best-defined phrase at hand, as far as I can tell.

faq.md
BatmanAoD: Well, so much for my "safe" phrasing above! But this is sort of what I was worried about and why I don't want to just say that the behavior is defined. Wasn't it pointed out in one of the RFC threads that LLVM with the MSVC linker currently _does_ generate code that causes `longjmp` to invoke destructors?

faq.md
BatmanAoD: Sorry, "it" here should have been "unwinding". You're right, this wasn't clear.

faq.md
BatmanAoD: Okay, that's true. I'll modify it to say that the only possible value that can be transferred is `c_int`.

faq.md
BatmanAoD: Hm, I actually just wasn't consistent about putting periods at the end of bullet points.

BatmanAoD:
>  ... claim that drop is always guaranteed to run in some platforms is true. 

I am guessing you are referring to the part where I used "it" ambiguously, and that my response there resolves this concern?

faq.md
gnzlbg: Sure, but what I'm saying is that the behavior of `setjmp` and `longjmp` in Rust is actually defined.

faq.md
gnzlbg: > Wasn't it pointed out in one of the RFC threads that LLVM with the MSVC linker currently does generate code that causes longjmp to invoke destructors?

Since for the case being discussed the behavior is undefined, running destructors is one of the many things that could happen. Whatever LLVM does can change any time as a consequence of this, and IIRC whether destructors are run or not for particular functions depends on things like the optimization level.

faq.md
gnzlbg: Note that you can use the return value of `setjmp` to transfer any kind of payload via the heap, just like we do for unwinding (the payload of `panic` is behind a `Box<dyn Any + ...>`), by doing a heap allocation before the `longjmp` and putting the pointer to it in a thread-local, that can be accessed from the `setjmp`.

faq.md
gnzlbg: So `setjmp` / `longjmp` have some requirements on how the language run-time works, e.g., the setjmp and longjmp must happen on the same thread, etc.

gnzlbg:
> I am guessing you are referring to the part where I used "it" ambiguously, and that my response there resolves this concern?

Yes, that resolves it.

PR #7, opened 2019-10-11
define roadmap readme etc

nikomatsakis:


roadmap/propagate-native-exception-through-Rust-frame.md
```diff
@@ -0,0 +1,39 @@
...
+
+* The Rust frame has **no in-scope destructors**.
+    * We need to define precisely when a Rust function is guaranteed to have no destructors.
+* The Rust frame **has destructors** it would like to execute.
+    * We need to define how they interact with a native exception.
+
+Currently, having a Rust panic "unwind" a native frame is **undefined
+behavior** in both of the above cases. However, we plan to specify the
+first case (no destructors) before we think about the more complex
+case (contains destrucors). We will also have to specify this on a
```
acfoltzer: typo `destrucors`

roadmap/propagate-native-exception-through-Rust-frame.md
```diff
@@ -0,0 +1,39 @@
...
+per-target basis, as the details will vary depending on what exception
+mechanism is in use.
```
acfoltzer: It might also be worth mentioning other variables. Perhaps:

> […] on what exception mechanism is in use, and what other non-exception features (such as `longjmp`) may use the same mechanism.

roadmap/propagate-native-exception-through-Rust-frame.md
```diff
@@ -0,0 +1,39 @@
...
+```
+     ...
+      |
+[Rust frame]
+      |                           ^
+      | (calls)                   | (unwinding
+      v                           |  goes this
+[native function]                 |  way)
+      |                           |
+      +--- rust function panics --+
```
acfoltzer: should be `native function panics`

roadmap/propagate-rust-panic-through-native-frame.md
```diff
@@ -0,0 +1,44 @@
...
+    * The precise meaning of "no destructors" is defined in terms of the native ABI.
+    * We need to define how the Rust panic propagates across such frames.
+* The native frame **has destructors** it would like to execute. These could come from C++
+  RAII, Java finally blocks, etc.
+    * We need to define how the Rust panic is defined and which destructors would execute.
+
+Currently, having a Rust panic "unwind" a native frame is **undefined
+behavior** in both of the above cases. However, we plan to specify the
+first case (no destructors) before we think about the more complex
+case (contains destrucors). We will also have to specify this on a
```
acfoltzer: typo `destrucors`

roadmap/propagate-rust-panic-through-native-frame.md
```diff
@@ -0,0 +1,44 @@
...
+```
+
+We can distinguish two cases:
+
+* The native frame has **no destructors**. An example of this might be a C function.
+    * The precise meaning of "no destructors" is defined in terms of the native ABI.
+    * We need to define how the Rust panic propagates across such frames.
+* The native frame **has destructors** it would like to execute. These could come from C++
+  RAII, Java finally blocks, etc.
+    * We need to define how the Rust panic is defined and which destructors would execute.
```
acfoltzer, line 36: It might be worth distinguishing a third case:

>* The native frame has **no destructors or any affordances for unwinding**. An example of this might be a C function compiled with `-fno-exceptions` or `-fno-unwind-tables`.
>    * The precise meaning of "affordances for unwinding" is defined in terms of the native ABI.

acfoltzer (commented):
In addition to these comments, my name should be listed as "Adam" in `README.md`. Thank you for writing all of this up!

roadmap/propagate-native-exception-through-Rust-frame.md
nikomatsakis: I debated whether to mention longjmp :) it seemed in the lang team meeting that people got real confusing about it. But I think it's worth discussing indeed.

roadmap/propagate-rust-panic-through-native-frame.md
nikomatsakis: ah, very nice

roadmap/propagate-native-exception-through-Rust-frame.md
acfoltzer: Yeah, there's definitely been a lot of confusion around it, but it does definitely need to be in consideration for Windows

README.md
```diff
@@ -16,3 +16,33 @@ support unwinding that crosses FFI boundaries.
 
 [shepherds-blog]: http://smallcultfollowing.com/babysteps/blog/2019/09/11/aic-shepherds-3-0/
 [zulip-room]: https://rust-lang.zulipchat.com/#narrow/stream/210922-wg-ffi-unwind/topic/welcome/near/177543226
+
+## Roadmap and current status
```
BatmanAoD, line 20: How would this section differ from `plan.md`? Is `plan.md` supposed to be for more "overarching" goals? (Or would we simply replace `plan.md` with the roadmap?)

roadmap/propagate-rust-panic-through-native-frame.md
```diff
@@ -0,0 +1,48 @@
...
+      +--- rust function panics --+
+```
+
+We can distinguish three cases:
+
+* The native frame has **no destructors**. An example of this might be a C function.
+    * The precise meaning of "no destructors" is defined in terms of the native ABI.
+    * We need to define how the Rust panic propagates across such frames.
+* The native frame **has destructors** it would like to execute. These could come from C++
+  RAII, Java finally blocks, etc.
```
BatmanAoD, line 35: This would only apply to Java code that is compiled rather than running in the JVM, right? It seems like an odd example, since (as far as I know) it's not possible for native exceptions to propagate into JVM stack frames (except by through the VM code itself).

roadmap/c-unwind-abi.md
```diff
@@ -0,0 +1,38 @@
...
+the target definition. In general, though, it is meant to match the
+"native" unwinding mechanism of the target: typically whatever C++
+compilers use.
+
+**Warning:** We are still in the process of fully specifying when and
+how "unwinding interop" works between native functions and Rust. This
+roadmap item alone **only** adds the ABI -- it does not define what
+should happen if unwinding actually occurs, and hence **even if you
+use this ABI**, unwinding across a `"C unwind"` ABI barrier is **still
+undefined behavior**. However, it is our **intent** to define that
```
BatmanAoD, line 22: Hm, do we want to call this "undefined behavior", without qualification? I had thought we intended (even in this initial minimal step) to at least specify enough of the behavior to prohibit `rustc` from emitting `nounwind` for the new ABI. As written, it would be an entirely valid implementation to treat `"C unwind"` as an alias for `"C"`.

roadmap/propagate-native-exception-through-Rust-frame.md
```diff
@@ -0,0 +1,43 @@
...
+      v                            |  goes this
+[native function]                  |  way)
+      |                            |
+      +--- native function throws -+
+```
+
+We can distinguish two cases:
+
+* The Rust frame has **no in-scope destructors**.
+    * We need to define precisely when a Rust function is guaranteed to have no destructors.
```
BatmanAoD, line 28: I'm not sure this is strictly true; I think it might actually be easier to specify the case with destructors (which is merely a superset of the case without destructors) without trying to define a "no-destructor" guarantee.

roadmap/propagate-rust-panic-through-native-frame.md
```diff
@@ -0,0 +1,48 @@
...
+```rust
+extern "C unwind" fn example() {
+    panic!("Uh oh");
+}
+```
+
+In this case, the stack would look like one of the two scenarios:
+
+```
+     ...
```
BatmanAoD, line 18: I think these two stack-frame diagrams are not really complete without showing that, to be well defined, the unwind must either be caught in the same language that threw it, or the application must terminate. Or, at least, I believe those are the only cases with which this working group should be concerned.

README.md
nikomatsakis: Probably we can just remove plan.md

roadmap/propagate-rust-panic-through-native-frame.md
nikomatsakis: Sorry, yeah, weird example, I was thinking of [gcj](https://en.wikipedia.org/wiki/GNU_Compiler_for_Java) but I guess that's not maintained anymore.

roadmap/c-unwind-abi.md
nikomatsakis: I say it is undefined behavior, full stop. It just happens to be behavior we **intend** to define. But at the moment, it is not defined. It could do anything (including what it actually does).

roadmap/propagate-native-exception-through-Rust-frame.md
nikomatsakis: Perhaps, yes. I guess we'll figure that out as we go. One thing that's not included in the doc is the order in which we do things per se -- we should probably write out the intended ordering, but I think it's something we should feel free to change.

roadmap/propagate-rust-panic-through-native-frame.md
nikomatsakis: Seems good.

roadmap/propagate-rust-panic-through-native-frame.md
BatmanAoD: I'm sure there's _some_ compiled language with `finally` blocks...I just don't know what it is! 😆 I'm guessing `rustc` targeting MSVC can't generate CLR code...

roadmap/c-unwind-abi.md
BatmanAoD: Hm, the oddity to me is that this is in a section that assumes the existence of `"C unwind"`, i.e., it is already defining new semantics. I'm not sure it's beneficial to introduce new semantics with no behavioral guarantees whatsoever.

PR #8, opened 2019-10-12
PR suggestions, terminology, links, etc

BatmanAoD:
This is fairly broad and could perhaps be broken down into multiple PRs. I believe it address all the comments raised on previous PRs.


roadmap/c-unwind-abi.md
```diff
@@ -21,7 +21,10 @@ should happen if unwinding actually occurs, and hence **even if you
 use this ABI**, unwinding across a `"C unwind"` ABI barrier is **still
 undefined behavior**. However, it is our **intent** to define that
 behavior in the future (and this is what the other roadmap items
-are all about). 
+are all about). Additionally, we consider unwinding across an `extern "C"`
+boundary to be [LLVM-undefined behavior][LLBM-UB], whereas for
```
BatmanAoD: This link is not rendering, but I'm not sure why.

roadmap/propagate-native-exception-through-Rust-frame.md
```diff
@@ -40,4 +44,27 @@ on a per-target basis, as the details will vary depending on what
 exception mechanism is in use, and what other non-exception features
 (such as `longjmp`) may use the same mechanism.
 
+[centril-effects]: https://github.com/Centril/rfc-effects/issues/11
+[FAQ-longjmp]: faq.md#how-does-cross-language-unwinding-differ-from-cross-language-setjmplongjmp
+
+## Unwind-termination and related concerns
```
BatmanAoD, line 50: These sections in the two files are almost identical. I am wondering whether having two separate files is the right approach.

I'm not yet sure of a better way to organize the information. I suspect a table might be good; I need to think more about what all the columns would be.

faq.md
```diff
@@ -1,31 +1,51 @@
...
+  considered undefined behavior). Cross-language unwind, however, will be
+  defined such that `Drop` objects whose frames are unwound are guaranteed
+  `drop`ed.
+- Unwinding across Rust frames when `panic = abort` is currently undefined
+  behavior, but we plan to define the behavior to cause the application to
+  `abort`. The behavior of `setjmp`/`longjmp`, however, is independent of the
+  `panic` runtime.
 - unwinding involves the use of a personality function, which raises additional
-  cross-language compatibility concerns; `setjmp`/`longjmp` does not
+  cross-language compatibility concerns; `setjmp`/`longjmp` does not.
```
gnzlbg, line 40: Doesn't `setjmp`/`longjmp` require a C-like stack ? For example, Go needs to copy its stack into a C-like stack before each FFI call, you can't call `longjmp`/`setjmp` from Python, etc.

roadmap/c-unwind-abi.md
gnzlbg:
```suggestion
boundary to be [LLVM-undefined behavior][LLVM-UB], whereas for
```

gnzlbg: This is undefined behavior in Rust, and that's what matters here. I don't see the value in making the distinction here that for a program for which we provide no guarantees broken LLVM-IR is generated, since we don't even guarantee that any LLVM-IR will be generated at all (and well, this would probably also break Cranelift, miri, and other rust backends).

roadmap/c-unwind-abi.md
```diff
@@ -36,3 +39,14 @@ you will have to keep up. Effectively, you're on a nightly release,
...
+
+## Panic = abort
+
+In order to safely call functions that may unwind, a Rust function must have
+a landing pad. Typically, `panic = abort` prevents landing pads from being
+generated, which would make the behavior of `"C unwind"`
+[LLVM-undefined][LLVM-UB]. To make this behavior well defined, we would like to
+require landing-pad generation for any function calling a `"C unwind"`
+function, even when compiling with `panic = abort`. These landing pads would of
+course `abort` the application rather than propagate the unwind.
```
gnzlbg, line 52: The only `extern "C unwind"` functions that can be safely called are defined in Rust and when `-C panic=abort` they never unwind so the case covered in this section cannot happen.

roadmap/propagate-native-exception-through-Rust-frame.md
```diff
@@ -40,4 +44,27 @@ on a per-target basis, as the details will vary depending on what
...
+  boundary
+
+Again, these are all currenntly undefined.
+
+We **do** expect to make the first and last possibilities (re-entering native
+frames or reaching the app entry point) well-defined; their behavior should be
+a natural consequence of how the `"C unwind"` ABI is defined.
 
+We **do not** currently have plans to define the middle case, wherein a native
+unwind propagates through Rust frames all the way to a thread boundary.
```
gnzlbg, line 70: What's the difficulty here?

In C++ `std::current_exception` returns NULL if the exception is "foreign", so if execution unwinds to a thread boundary, and you can't get a pointer to the current exception, then that's an exception that you cannot propagate to another thread, e.g., because it is a "foreign" exception, and you can just `abort` the program.

roadmap/propagate-rust-panic-through-native-frame.md
```diff
@@ -46,3 +48,37 @@ cases (contains destructors, no affordances). We will also have to
...
 on what exception mechanism is in use.
 
+Another consideration is whether the the native frames include a `catch` block.
+There are several possibilities here as well:
+
+* "catch native" -- e.g. a C++ block catching a specific subset of C++ exceptions
+* "catch all" -- e.g. a C++ catch without an exception argument
+* "catch and release" -- e.g. a C++ "catch all" that re-`throw`s the exception
+
+So far we do **not** have plans to specify these cases.
```
gnzlbg, line 58: Nor could we if we wanted. The behavior of these are up to the C++ language / implementation, and nothing that we would specify here would change that, or prevent it from changing. At most, we could document what the behavior is, but that's different from specifying it.

roadmap/propagate-rust-panic-through-native-frame.md
```diff
@@ -46,3 +48,37 @@ cases (contains destructors, no affordances). We will also have to
...
+
+## Unwind-termination and related concerns
+
+There are several possibilities for the top section of the diagram, above the
+native frames:
+
+* Rust frames - the `panic` re-enters Rust frames
+  * At this point, the `panic` will be no different from a normal `panic`,
+    unless and until it re-enters native frames.
```
gnzlbg, line 67: All other points in this list follow from this one for Rust at least.

roadmap/propagate-rust-panic-through-native-frame.md
```diff
@@ -46,3 +48,37 @@ cases (contains destructors, no affordances). We will also have to
...
+  * The `panic` may caught as usual, either with `catch_panic` or by crossing a
+    thread boundary.
+* Thread boundary - the exception propagates all the way to a native frame that
+  was invoked from another thread
+* App entry point - the unwind is not caught and does not cross a thread
+  boundary.
+  * This should cause the application to terminate.
+
+Again, these are all currenntly undefined.
```
gnzlbg, line 76: The only thing that would make any of these undefined is the clause above that says that "currently having a Rust panic unwind a native frame is UB". But that clause makes unwinding through pure Rust code UB, and in pure Rust code, all of the above is defined. 

That clause also makes the "C unwind" ABI no different in practice from the "C" ABI, and allows us to emit `nounwind` for `"C unwind"` since a program that would actually unwind with that ABI would have UB.

spec-terminology.md
```diff
@@ -0,0 +1,30 @@
...
+behavior:
+
+#### LLVM-undefined behavior (or LLVM-UB)
+
+Cases of known LLVM-UB are a specific subset of undefined behavior in general.
+Rust code with LLVM-UB will cause `rustc` to generate LLVM IR exhibiting
+undefined behavior.
+
+This is distinct from the general case of Rust undefined behavior, in which
+it is unknown whether `rustc` will generate well-behaved LLVM IR.
```
gnzlbg, line 30: I found calling this "LLVM-UB" confusing because it sounds as if this is only UB in LLVM but not in Rust, while this are particular instances of undefined behavior in Rust that are known to generate LLVM-IR that has undefined behavior in LLVM as well.

roadmap/propagate-rust-panic-through-native-frame.md
```diff
@@ -46,3 +48,37 @@ cases (contains destructors, no affordances). We will also have to
 specify this on a per-target basis, as the details will vary depending
 on what exception mechanism is in use.
```
gnzlbg, line 49: This paragraph and the bullet points at the end, ignore the fact that native frames can be written in Rust. That is, the call graph above can be Rust->Rust->Rust, and this then specifies that unwinding with a Rust panic through Rust is UB. 

The bullet points at the end of the file also ignore this, making things like the behavior of a Rust panic crossing a Rust thread boundary undefined.

roadmap/propagate-rust-panic-through-native-frame.md
```diff
@@ -46,3 +48,37 @@ cases (contains destructors, no affordances). We will also have to
...
+
+Again, these are all currenntly undefined.
+
+We **do** expect to make the first and last possibilities (re-entering Rust
+frames or reaching the app entry point) well-defined; their behavior should be
+a natural consequence of how the `"C unwind"` ABI is defined.
+
+We **do not** currently have plans to define the middle case, wherein a `panic`
+propagates through native frames all the way to a thread boundary.
+
```
gnzlbg, line 84: So this is a meta-comment about the whole file. I think this is the wrong approach to attack this problem.

The whole point of an ABI is to specify an interface. We have to specify whether Rust conforms to the interface or not, and if it does, how Rust behaves at its side of the interface. 

We could document how other native code specifies their behavior at their side of the interface, but we can't specify such behavior for that code, and we can't guarantee that such behavior does not change. 

Specifying here that if a foreign exception crosses a catch block in C++ the behavior is undefined is as meaningful as C++ specifying that integer arithmetic is undefined behavior in Rust. Sure, that's something that C++ could write in their language spec, but that's neither true nor useful.

For example, for all the cases considered here about what happens w/o destructors, catch blocks, thread boundaries, etc. C++ implementations actually specifies the behavior of foreign exceptions that adhere to the platform ABI, so the only way in which that could result in the UB specified here would be if the "C unwind" feature would not have the ABI that it says it does, which would make the feature unsound.

gnzlbg: So instead of writing what we think should be UB or not at the other side of the FFI, I'd very much prefer if we would just write what we instead _guarantee_ for our side, and what we guarantee for and require from the interface. These guarantees and requirements are what allow interfacing with native code correctly, since they tell this code what it can rely on, and what things it must uphold to avoid UB on our side. To avoid UB on its side, such code should refer to the specification of whatever language it is written in.

Since we have already said that functions with the `"C unwind"` ABI unwind with the platform ABI, Rust definitions with that ABI must uphold that guarantee, and so must functions that are declared in Rust with that ABI but defined in unknown code (independently of whether that code is Rust or not).

For the Rust side we need to define the behavior for the two cases:

* the unwinding is due to a Rust panic: in this case, it behaves like a normal Rust panic, for which we already have defined the behavior - anything else would break Rust->Rust unwinding through "C unwind".
* the unwinding is due to a "foreign" exception (e.g. a C++ `throw`): we can initially say that the behavior is undefined (which can always be defined later)

That would suffice for an MVP that allows Rust->C++->Rust to work correctly with strong stability guarantees for the targets for which the `"C unwind"` ABI is available.

faq.md
BatmanAoD: The point is that there are no cross-language compatibility concerns _due to personality functions_, not that `setjmp`/`longjmp` work in all languages.

roadmap/c-unwind-abi.md
BatmanAoD: This is discussing a future in which it's safe to call functions that _can_ unwind using `extern "C unwind"`. (Even if the function is in Rust, it could, for instance, be in a shared library compiled with `panic = unwind`.)

roadmap/propagate-native-exception-through-Rust-frame.md
BatmanAoD: > In C++ `std::current_exception` returns NULL if the exception is "foreign"...

That's not guaranteed by [the relevant section of the standard](https://timsong-cpp.github.io/cppwp/propagation#lib:current_exception).

Another issue is that `panic` is translated to an `Err` value at thread boundaries, but as far as I can tell, doing that translation for an exception would require catching the exception in Rust, which we don't currently plan to do.

faq.md
gnzlbg: Makes sense. 

Yes, I think `longjmp` doesn't have those even though they can unwind on some platforms.

gnzlbg: Maybe this is clearer? 

```suggestion
  cross-language compatibility concerns due to personality functions; `setjmp`/`longjmp` does not.
```

roadmap/propagate-rust-panic-through-native-frame.md
BatmanAoD: The reason I stated it this way is that the behavior of foreign code interacting with a Rust `panic` is of course dependent on parts of the `panic` ABI that otherwise do not need to be specified.

I mentioned this in Zulip, but I believe one such ABI element is the personality function. I don't believe any other verbiage in this document so far implies any constraints on the personality function, so I think it's good to be explicit here.

roadmap/propagate-native-exception-through-Rust-frame.md
gnzlbg: No, it is a platform guarantee. Itanium guarantees that.

roadmap/propagate-rust-panic-through-native-frame.md
BatmanAoD: I'm not sure what you mean - are you saying that the thread- and app- boundary cases don't need to be stated?

roadmap/propagate-rust-panic-through-native-frame.md
gnzlbg: > The reason I stated it this way is that the behavior of foreign code interacting with a Rust panic is of course dependent on parts of the panic ABI that otherwise do not need to be specified.

The Rust panic ABI does not interact with whatever unwinding ABI "C unwind" uses. At the "C unwind" boundary we translate Rust panics into a different kind of panic, e.g.,  a "platform Rust panic", and foreign code only interacts with those, not with Rust panics themselves.

roadmap/propagate-rust-panic-through-native-frame.md
BatmanAoD: > That clause also makes the "C unwind" ABI no different in practice from the "C" ABI...

No, it makes the ABIs no different in _specification_ (at least temporarily). In practice, we won't actually do this:

> and allows us to emit nounwind for "C unwind" since a program that would actually unwind with that ABI would have UB.

...but we will continue to do so for `"C"`. And, furthermore, we will not leave the `"C unwind"` behavior undefined indefinitely.

roadmap/propagate-rust-panic-through-native-frame.md
gnzlbg: The only thing that needs to be possible is this translation, but this is always possible AFAICT, because we can always use `catch_unwind` with Rust panics, and after catching the panic, we can do whatever we want with it to make it conform to the native ABI.

roadmap/propagate-rust-panic-through-native-frame.md
BatmanAoD: "Native" is intended to mean "not Rust". I'll go through and clarify that; thank you.

roadmap/propagate-rust-panic-through-native-frame.md
gnzlbg: More like all these guarantees can be stated by just saying that: "When an `extern "C unwind" { fn foo(); }` declaration unwinds and the unwinding was started with a Rust panic, that unwinding proceeds exactly as if it would have happened in Rust". 

This means that all other bullet points then just follow from how `panic!` is currently guaranteed to work in Rust.

roadmap/propagate-rust-panic-through-native-frame.md
BatmanAoD: > C++ implementations actually specifies the behavior of foreign exceptions that adhere to the platform ABI...

This is conflating two kinds of specification. The C++ _standard_ specifies no such thing. What we are writing in these documents are more akin to a Rust _standard_ than to an _implementation_ specification. So it is just as reasonable for us to leave these things undefined as it is for C++ to do so.

We do want Rust _implementations_ supporting `extern "C unwind"` to specify their behavior, but this is a bit more of a long-term goal than simply omitting `nounwind`, which is why we have all of the future-facing verbiage, e.g. "we plan to define the behavior".

roadmap/c-unwind-abi.md
gnzlbg: If calling an `extern "C unwind"` function cannot unwind, then this feature adds no value. For this feature to add any value to the language, such functions must be allowed to unwind.

roadmap/propagate-rust-panic-through-native-frame.md
gnzlbg: > No, it makes the ABIs no different in specification (at least temporarily).

Sure, I do not think that an RFC that proposes this can obtain any kind of consensus. Because then this feature introduces real complexity to the spec without adding any value, making it impossible to analyze the tradeoffs. All analysis would need to be hypothetical referring to potential future value that the feature might add. I personally would strongly suggest to defer the RFC until such value can be delivered (note that, implementation wise on nightly, the bar is much lower).

roadmap/propagate-rust-panic-through-native-frame.md
gnzlbg: > What we are writing in these documents are more akin to a Rust standard than to an implementation specification.

RFCs are not for a Rust standard decoupled from an implementation. They are for the Rust toolchain as a whole: the Rust reference, the Rust compiler, cargo, etc.

BatmanAoD: @gnzlbg I'm not sure how true that is, but this isn't even an "RFC" right now.

gnzlbg: Isn't this material that should lead to an RFC at some point ?

BatmanAoD: Yes, but for now let's just assume that the material in the roadmap is at "language specification" level.

faq.md
```diff
@@ -1,31 +1,51 @@
...
-- It should never be assumed that `drop` will be called for objects in
-  intermediate frames traversed by a `longjmp`, but this may occur on certain
-  platforms. Rust provides no guarantee either way. Conversely, on platforms
-  where unwinding is supported, it will always call `drop`.
+
+### How does cross-language unwinding differ from cross-language `setjmp`/`longjmp`?
+
+- `setjmp`/`longjmp` across Rust frames is currently guaranteed to have
+  well defined behavior as long as those frames do not contain destructors.
+  When crossing frames that do contain deestructors, the  behavior of `longjmp`
```
nikomatsakis, line 22:
```suggestion
  When crossing frames that do contain destructors, the  behavior of `longjmp`
```

faq.md
```diff
@@ -1,31 +1,51 @@
...
-- `setjmp`/`longjmp` across Rust frames is currently guaranteed not to have
-  undefined behavior or to `abort`
-- It should never be assumed that `drop` will be called for objects in
-  intermediate frames traversed by a `longjmp`, but this may occur on certain
-  platforms. Rust provides no guarantee either way. Conversely, on platforms
-  where unwinding is supported, it will always call `drop`.
+
+### How does cross-language unwinding differ from cross-language `setjmp`/`longjmp`?
+
+- `setjmp`/`longjmp` across Rust frames is currently guaranteed to have
```
nikomatsakis, line 20: do we have a citation for this? that's a strong term :) though I think it is reasonable.

roadmap/propagate-rust-panic-through-native-frame.md
nikomatsakis: I'm going to mark this as "resolved", but I do think that ultimately the right way for us to proceed is going to specifying *what we guarantee as part of the "C unwind" ABI* -- and then discussing the implications of those guarantees

nikomatsakis (approved):

faq.md
gnzlbg: Since Rust itself cannot call them I don't think we have wording anywhere explicitly guaranteeing that this works. According to the requirements and guarantees we provide, when a Rust frame is "sandwiched" across C frames, and C does that, then as long as the frame has no types with destructors, none of the documented types of UB that we have in the reference can happen (https://github.com/rust-lang-nursery/reference/blob/master/src/behavior-considered-undefined.md), although that list is incomplete.

We have gone to certain extends to make sure that this works for, e.g., rlua, making sure that longjmps of lua across Rust do not cause Rust to abort, even for `nounwind` functions.

nikomatsakis:
OK, I'm going to merge this PR, because I'd prefer to work iteratively. The remaining things I think we could address that I recall:

* don't talk about *undefined behavior* but rather *to be determined* behavior
* see if we can rephrase the some of the "roadmap" stages to talk about the things we do (or, at least to start, mostly do not) guarantee

PR #9, opened 2019-10-15
[discussion-only] My thoughts an MVP RFC for the full stable "C unwind" feature

gnzlbg:
So this is what I would consider an MVP for stabilization.

### DIFF
--- a/rfcs/0000-simple_unwind_annotation.md
+++ b/rfcs/0000-simple_unwind_annotation.md
@@ -280,7 +280,23 @@ TODO
 # Unresolved questions
 [unresolved-questions]: #unresolved-questions
 
-TODO
+## Lossless Rust panics through `"C unwind"`
+
+The text provides the following guarantee:
+
+> When a panic that originates in Rust, unwinds
+> through a `"C unwind"` function call back into Rust, the translation is
+> guaranteed to be lossless _if the panic was not modified by non-Rust code_, and
+> this then behaves as if the panic would have never left Rust.
+
+I'm not sure if we can uphold it on all targets that where `"C unwind"` makes
+sense, but we can provide it on all Itanium-like and probably the windows
+targets, so it might be a worth trade-off to just say that targets that `"C
+unwind"` is not available on targets that cannot support this.
+
+We could always add a different ABI string with weaker guarantees if Rust were
+to ever support any such target. This feature doesn't need to solve everybody's
+problems.
 
 # Future possibilities
 [future-possibilities]: #future-possibilities
--- a/rfcs/0000-simple_unwind_annotation.md
+++ b/rfcs/0000-simple_unwind_annotation.md
@@ -174,13 +174,15 @@ panics crossing function boundaries with the `"C unwind"` ABI are translated
 back and forth from the Rust unwinding ABI to the "native Rust panic ABI" which
 comforms to the platform ABI. This translation might incur a cost on some
 platforms, and this cost might change over time as the Rust panic ABI evolves.
+If this were to happen, the ["Future Possibilties"][future-possibilities]
+section of this RFC has you covered.
 
-`"C unwind"` functions can unwind:
+In contrast with `"C"`, `"C unwind"` functions can unwind.
 
-* if the `"C unwind"` function is defined in Rust, it unwinds the stack, and the
+* If the `"C unwind"` function is defined in Rust, it unwinds the stack, and the
   panic can be caught with `catch_unwind`.
   
-* if the `"C unwind"` function is not defined in Rust, it unwinds the stack. 
+* If the `"C unwind"` function is not defined in Rust, it unwinds the stack. 
   * whether such unwindings can always, sometimes, or never be caught with
   `catch_unwind` or not is target-dependent. If some of these unwindings that do
   not originate in Rust can be caught, their value is then of type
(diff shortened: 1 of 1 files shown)


rfcs/0000-simple_unwind_annotation.md
```diff
@@ -1,179 +1,311 @@
...
+more targets in the future, either in other RFCs that build on this one, or in
+FCPs.
+
+Like for all other ABI strings, one can use this ABI string:
+
+  * to define a function in Rust: `extern "C unwind" fn foo() { }`
+  * to declare a function in Rust: `extern "C unwind" { fn foo(); }`
+  * in function pointer types: `let _: extern "C unwind" fn = foo;`
+    * Since `"C unwind"` is a super-set of `"C"`, there are implicit coercions
+      from `extern "C"` function types to `extern "C unwind"` function types.
```
nikomatsakis: I would prefer not to have such coercions, in part just to keep things simple. I think they are a fine bit of potential future work, though. That said, I guess it wouldn't be too hard to implement -- I wasn't sure if we would want to guarantee that "C" and "C unwind" are interoperable. Probably we simply do, in which case such coercions seem harmless enough.

rfcs/0000-simple_unwind_annotation.md
```diff
@@ -1,179 +1,311 @@
...
+      These are analogous to the coercion rules for safe `fn` types to `unsafe
+      fn` types.
+
+Similarly to the `"Rust"` ABI string, `"C unwind"` functions can unwind:
+
+* if the `"C unwind"` function is defined in Rust, it unwinds the stack, and the
+  panic can be caught with `catch_unwind`.
```
nikomatsakis: I presume you mean *and the panic originated in Rust* or something like that?

rfcs/0000-simple_unwind_annotation.md
```diff
@@ -1,179 +1,311 @@
...
+  
+* if the `"C unwind"` function is not defined in Rust, it unwinds the stack, but
+  whether such unwindings can always, sometimes, or never be caught with
+  `catch_unwind` or not is target-dependent. If some of these unwindings that do
+  not originate in Rust can be caught, their value is then of type
+  `std::panic::ForeignException` (that is, the `Result::Err(Any)` that
```
nikomatsakis, line 146: I feel like we should leave "catching of foreign exceptions" for future work and not try to say anything about it.

rfcs/0000-simple_unwind_annotation.md
```diff
@@ -1,179 +1,311 @@
...
+  `catch_unwind` returns can be downcasted to such a type).
+
+The Rust panic ABI is unspecified. When a `"C unwind"` function that is defined
+in Rust unwinds, the Rust panic is translated to a "native Rust panic" which
+conforms to the native ABI. When a panic that originates in Rust, unwinds
+through a `"C unwind"` function call back into Rust, the translation is
+guaranteed to be lossless _if the panic was not modified by non-Rust code_, and
+this then behaves as if the panic would have never left Rust.
```
nikomatsakis, line 154: I included similar text in my draft RFC.

nikomatsakis (approved):
This seems pretty close to the "Draft RFC" I was working on. At some future point I'd like to take some of this material. Certain details I think we may wish to defer (notably, the details about catching non-Rust exceptions).

rfcs/0000-simple_unwind_annotation.md
gnzlbg: I agree. I don't think the coercions are required for an MVP, I just had the rules already written somewhere and c&p them here. I think I'm going to move this to the future section. If someone were to implement this before the PR for stabilization, one can re-consider at that point whether we want to stabilize them as well or leave them as unstable, but they do not need to block the stabilization of this feature.

rfcs/0000-simple_unwind_annotation.md
gnzlbg: Indeed, thanks.

rfcs/0000-simple_unwind_annotation.md
gnzlbg: What I'm not sure is if this "goal" is achievable everywhere (it is on the Itanium-like targets, and probably on Windows). 

I worded it as a guarantee because I think we should really try to guarantee this because I have the feeling that otherwise weird things might happen but I don't know what they would be. 

If most targets can support this, it might be worth it to consider whether it would be acceptable for `"C unwind"` to just not be available on targets in which this cannot be upheld. We can always offer a different ABI string with weaker guarantees for those.

I'm going to add this as the first unresolved question.

rfcs/0000-simple_unwind_annotation.md
gnzlbg: This is a complex part of the `"C unwind"` feature so I'm not sure either if it belongs in an MVP. 

I think that if we were to say that "unwinding with a foreign exception into Rust is not allowed (UB)", then most of the complexity goes out.

What brings most of the complexity is allowing foreign exceptions to unwind into Rust, e.g., to support `longjmp`, or `pthread_cancel`, because then it would be nice if we at least document what that does: whether unwinding calls destructors, whether `catch_unwind` aborts, or catches, or lets foreign exceptions escape, whether `catch_unwind` can distinguish Rust panic from foreign exceptions, etc.

If that's a complexity we want to accept for an MVP, then being able to catch foreign exceptions adds little extra complexity, since C++ has done all of the design and implementation work for us with `std::exception_ptr`, and clang already implements it for the same targets that we support, while at the same time adding a lot of power for multi-threaded programs. For example, without catching foreign exceptions, when a foreign exception reaches a thread boundary the program aborts or has UB. But if you can catch them, you can actually pass that exception to the calling thread and then re-raise it and continue unwinding to fail gracefully, or just log it and directly re-start the thread, etc. 

I can move this feature to the future section, but I think that as we investigate what the targets can and cannot do, we should, since we are already at it, document how foreign exceptions work or can work on each target.

BatmanAoD:
Superseded by #28

PR #10, opened 2019-10-16
Ub vs tbd

nikomatsakis:


faq.md
```diff
@@ -17,19 +17,22 @@
...
 
-- `setjmp`/`longjmp` across Rust frames is currently guaranteed to have
-  well defined behavior as long as those frames do not contain destructors.
-  When crossing frames that do contain deestructors, the  behavior of `longjmp`
-  is undefined; conversely, a primary goal of defining cross-language unwinding
-  behavior is to support crossing frames with destructors.
+- `setjmp`/`longjmp` across Rust frames is currently intended to have
+  well defined behavior as long as those frames do not contain
+  destructors, although we don't have any documentation to that
+  effect.
```
gnzlbg: I've opened an UCG issue about this, maybe it would be just better to link there: https://github.com/rust-lang/unsafe-code-guidelines/issues/210 

Some thoughts in the issue might influence other parts of this section.

roadmap/c-unwind-abi.md
```diff
@@ -3,7 +3,7 @@
 ## Summary
 
 Functions that use the plain `"C"` ABI are **not permitted to unwind**.
-Doing so is "undefined behavior", which means that the compiler is free
+Doing so is [Undefined Behavior], which means that the compiler is free
```
gnzlbg, line 6: I'd remove the part after the comma, it should be clear for the definition of UB that this is the case.

roadmap/c-unwind-abi.md
```diff
@@ -16,30 +16,24 @@ compilers use.
...
-which **likely will change** across Rust stable releases. Therefore,
-you will have to keep up. Effectively, you're on a nightly release,
-even though you're using only stable syntax. (The same is true for
-many aspects of unsafe code.)
-
-[LLVM-UB]: ../spec-terminology.md#LLVM-undefined-behavior-or-LLVM-UB
+roadmap item alone **only** adds the ABI string to Rust. All details
+of how that ABI strings are implemented for various targets are still
+considered [To Be Defined]. As a result, programs that rely on those
+details are not truly stable; you may find that the details change as
```
gnzlbg, line 22: This suggests that we would allow such "To Be Defined" behavior in stable Rust, but understood the intent to be to define this behavior before stabilization.

roadmap/c-unwind-abi.md
```diff
@@ -16,30 +16,24 @@ compilers use.
...
+
+## The goal
+
+In practice, there are a number of crates that rely on unwinding
+across a "C" ABI today -- even though that is [Undefined
+Behavior]. Creating the "C unwind" ABI means that those crates can
+migrate to this ABI and better express their intent.  It does not (in
+and of itself) make the behavior of those crates defined -- they are
+still relying on [To Be Defined] details. But it *does* mean that they
+will no longer trigger [Undefined Behavior].
```
gnzlbg, line 36: We have crates using `"C"` to interface with foreign code that exhibit UB because, among other things, the requirements that this foreign code places on Rust are not guaranteed by Rust. The goal here is stated as offering these crates a `"C unwind"` feature that does not guarantee these requirements either. I don't understand how this is no longer UB. 

For this to no longer be UB, we would need to guarantee that what that foreign code requires is upheld (i.e. that we implement the ABI that it uses correctly), yet we don't intend to do that, so, I'm confused.

spec-terminology.md
```diff
@@ -1,30 +1,75 @@
...
 
-* Well-defined behavior
-* Implementation-defined behavior
-* Unspecified behavior
-* Undefined behavior
+## Undefined Behavior
 
-Of these, only "undefined behavior" is used consistently within the Rust
-project; there are _no_ guarantees about the runtime behavior of such code.
+As is typical within the Rust community we use the phrase **undefined
```
gnzlbg, line 13: I think it would be better to just link here to the definition in the UCG repo glossary.

spec-terminology.md
```diff
@@ -1,30 +1,75 @@
...
+<a name="LLVM-UB"></a>
 
-Cases of known LLVM-UB are a specific subset of undefined behavior in general.
-Rust code with LLVM-UB will cause `rustc` to generate LLVM IR exhibiting
-undefined behavior.
+### LLVM-undefined behavior (LLVM-UB)
 
-This is distinct from the general case of Rust undefined behavior, in which
-it is unknown whether `rustc` will generate well-behaved LLVM IR.
+As a special case of undefined behavior, we use the phrase **LLVM
```
gnzlbg: It is unclear whether LLVM-UB is also UB in Rust.

spec-terminology.md
```diff
@@ -1,30 +1,75 @@
...
+
+We refer to some behavior as **to be defined** to indicate that --
+while it is currently unspecified -- we *intend* to define that
+behavior at some point as part of this project group (though plans can
+change). This helps to define the scope of the group, but it also
+indicates behavior that you would be able to rely upon in the future
+in stable code. Note that TBD behavior is **still unspecified** until
+a formal decision is made, though, so if you rely on it today, your
+code cannot be considered stable Rust (even if it compiles on the
+stable compiler).
```
gnzlbg: So my idea of TBD is that it is a term that is used within this WG only, where we use it to say that we haven't given the issue the necessary attention to specify things in any way but have to do so at some point. So in practice, it is just undefined behavior. 

I also think that TBD should not reach "Rust", that is, everything that's TBD needs to be specified to be some type of behavior before stabilization (including just saying that its UB). 

Why use it instead of just UB? As a placeholder to differentiate the UB that's "UB by design" from the UB that's "UB because we haven't given this any thought yet, but want to do so".

gnzlbg (commented):
There is an issue open in the UCG repo about defining these types of behavior that maybe we should pursue more actively there (https://github.com/rust-lang/unsafe-code-guidelines/issues/201 https://github.com/rust-lang/unsafe-code-guidelines/issues/202). I think that in the meantime, it's fine to add things here, but we should update this to point there as things achieve consensus (for UB, this already happened).

roadmap/c-unwind-abi.md
nikomatsakis: I think I clarified the wording now.

spec-terminology.md
nikomatsakis: > So my idea of TBD ...

I think this basically aligns with what I wrote. I can't tell if there is a suggested edit in there? If so, can you try to make it more concrete?

> I also think that TBD should not reach "Rust", that is, everything that's TBD needs to be specified to be some type of behavior before stabilization (including just saying that its UB).

You and I have a different opinion here. I agree that in general it would be *nice* if TBD behavior remains on nightly, but I think it is acceptable for it to hit stable as well, and this has already happened in all sorts of areas -- the most notable case is the interaction of raw pointers, but there are plenty of other examples (e.g., the details of symbol visibility and shraed libraries is something we've tinkered with a few times). In general, I think it is a strict improvement on the status quo to offer people a plan they can adopt and a clear definition of what they can and cannot rely upon (and make progress on the latter until it suffices). However, I would be happy to try and write-up a "decision report" and take it to the lang team for a more formal decision.

spec-terminology.md
nikomatsakis: I think that the phrase "special case" implies that it is. However, I'm not sure how important the distinction really is at the end of the day.

That said, I debated about whether to include that phrase. I think it's probably worth distinguishing LLVM-UB as an orthogonal concept -- LLVM-UB is whatever LLVM defines as UB. Any case where Rust generates LLVM that exhibits LLVM-UB is a bug unless that behavior is also UB in Rust -- but those bugs can and do happen so it's good to be able to talk about it.

roadmap/c-unwind-abi.md
nikomatsakis: I don't really understand what is confusing. Presumably that code is relying on some detail that is not yet specified. This means that when it runs, something will obviously happen, but what precisely that something is can change from rust version to rust version or with other vagaries. 

In short, relying on TBD or unspecified details **is** a form of UB, in my opinion. The difference is *very* slight.

> For this to no longer be UB, we would need to guarantee that what that foreign code requires is upheld (i.e. that we implement the ABI that it uses correctly), yet we don't intend to do that, so, I'm confused.

What do you mean by "we don't intend to do that"?

roadmap/c-unwind-abi.md
nikomatsakis: I'm going to leave it because I want people who don't click through the link and are not that familiar to understand the context. I agree it is not strictly required.

spec-terminology.md
gnzlbg: > I think that the phrase "special case" implies that it is. 

I thought that the goal of the term "LLVM-UB" was to use it for those cases in which Rust has UB, but we guarantee that the generated LLVM-IR does not. For example, unwinding through "C unwind" is Rust-UB because we don't want to guarantee what that does, but it is not LLVM-UB because we want people to use it. If "LLVM-UB" implies "Rust-UB" then by saying that "C unwind" is not "LLVM-UB" we are saying that it is not "Rust-UB". Does that make sense ?

spec-terminology.md
gnzlbg: > You and I have a different opinion here. I agree that in general it would be nice if TBD behavior remains on nightly, but I think it is acceptable for it to hit stable as well, and this has already happened in all sorts of areas -- the most notable case is the interaction of raw pointers, but there are plenty of other examples (e.g., the details of symbol visibility and shraed libraries is something we've tinkered with a few times).

I think the issue is that by TBD you refer to what I refer to as "unspecified behavior", where we say that this behavior is defined to do something that's not UB, but we don't say what that something is.

Right now, almost everything is unspecified in these documents, but some things are "unspecified by design" - we have looked at them, and this is something that we have decided that we don't want to specify, like the Rust unwinding ABI - and some are "unspecified because we haven't had the time to look at it yet". 

I was hoping that TBD could be used for these second type of unspecified things, so that we would at least look at them before stabilization, and say whether we want them to be undefined, unspecified, or something else. 

Maybe from this point-of-view it is clearer why I think that shipping a TBD is not a great idea. It would mean that we have recognized that there is something important enough to mention here as something to consider, but that we haven't considered it enough to even say that we provide no guarantees about it.

roadmap/c-unwind-abi.md
gnzlbg: See below, I think that by TBD I was referring to some other thing, so probably we should clarify that first.

spec-terminology.md
nikomatsakis: So I think the full range of things are:

* Undefined behavior -- always a bug. As we discussed some in privmsg, the "miri virtual machine" panics here.
* Unspecified behavior -- the machine does not panic, but we don't say what happens. As you point out, there are a number of *reasons* we might not have said what happens yet:
    * We never want to specify this because we expect to want to change it. Example: Rust struct layout (without a repr attribute).
    * We expect eventually to specify this, but we haven't figure out just how it should be specified *yet*. This is what I am calling TBD, and I think you are too.

An orthogonal discussion is whether "TBD" behavior should ever be possible on stable. It seems clear that this is *suboptimal*, in that it leaves room for "de facto" dependencies. 

I have tried to summarize this in a [hackmd doc](https://hackmd.io/7ZcKCaRiRO2_I_yK8yeauQ).

nikomatsakis: Resolving this concern based on [this comment](https://rust-lang.zulipchat.com/#narrow/stream/210922-wg-ffi-unwind/topic/tbd.20vs.20ub.20vs.20unspecified.20behavior/near/178398565).

README.md
```diff
@@ -7,7 +7,7 @@ support unwinding that crosses FFI boundaries.
   - [acfoltzer (Adam)](https://github.com/acfoltzer)
   - [batmanaod (Kyle)](https://github.com/batmanaod)
 - Rust lang team contacts:
-  - [nikmoatsakis (Niko)](https://github.com/nikmoatsakis)
+  - [nikomatsakis (Niko)](https://github.com/nikmoatsakis)
```
BatmanAoD, line 10:
```suggestion
  - [nikomatsakis (Niko)](https://github.com/nikomatsakis)
```

My bad!

spec-terminology.md
nikomatsakis: I think this concern is basically resolved now -- I've clarified that LLVM-UB just refers to what is UB in LLVM, which is (hopefully) a subset of Rust UB (but not necessarily).

Moreover, as per this document, unwinding through "C unwind" is not Rust UB, but it *is* unspecified (or, more precisely, "TBD").

nikomatsakis: Resolving.

faq.md
BatmanAoD: Alex Crichton was the one who told me that the intention is for `setjmp`/`longjmp` to be "well-defined" as long as there aren't `impl Drop` objects in the frames crossed. His exact words were "you can always longjmp across rust frames". Maybe we just need to get that guarantee into the Reference itself?

Should we ask Alex if the guarantee is "strong" enough to be added to the Reference?

roadmap/c-unwind-abi.md
BatmanAoD: I think the core problem here is that Undefined Behavior has a very well-defined technical definition, but large parts of the C++ community (which I expect has the largest mind-share for the term) interprets it almost maliciously. (I actually love the quote "a sufficiently advanced compiler is indistinguishable from an adversary.") So the idea that something is "undefined but useful in practice", while historically valid (and, I'm told, part of the original intent for leaving some specific bits of C undefined), seems to be considered a contradiction in terms by some (I remember but cannot find a quote from Stack Overflow saying that UB can "create a black hole inside your cat").

So somehow we need to convey "the Rust compiler will not interpret `extern "C unwind"` adversarially", but that's not very clear or even terribly reassuring.

resolved-concerns/2019-10-17-tbd-on-stable.md
```diff
@@ -0,0 +1,44 @@
...
+        * Example: whether some local variable is materialized on the stack.
+    * We *may* specify this at some point in the future, but there are no plans to do so.
+        * Example: Rust ABI compatibility
+        * Example: What symbols get exported by a DLL
+    * We wish to specify this behavior in the near future, as part of the FFI-unwind project
+        * This is what I am calling **to be determined behavior**, but this is not a term that would appear in the Rust reference. In that context, we would describe such behavior as unspecified, but link to the ffi-unwind project as a place where users can learn more.
+        * Examples:
+            * Details of how a Rust panic presents itself in "C unwind" ABI on msvc
+
+## TBD as an "project-local planning measure"
```
BatmanAoD, line 18:
```suggestion
## TBD as a "project-local planning measure"
```

faq.md
gnzlbg: It might be better to put the guarantee through the RFC process. @nikomatsakis what do you think? 

I think that such an RFC could be really small. All the complications arise from either considering frames with destructors and/or trying to allow users to call `setjmp` from Rust (which returns multiple times, allows stopping execution in the middle of a Rust frame, etc.). 

Without that, the only thing that the RFC has to add is the guarantee that deallocating a rust frame that does not contain any drop glue is equivalent to `mem::forget`ing all the automatic variables of that frame. 

The hardest question would probably be "where in the reference do we put such a guarantee" (EDIT: maybe in the function items section since this is about function frames).

resolved-concerns/2019-10-17-tbd-on-stable.md
```diff
@@ -0,0 +1,44 @@
...
+
+* Undefined Behavior -- always a bug. Conceptually, the "miri virtual machine" panics here. This is the same definition [in use by the unsafe code guidelines](https://rust-lang.github.io/unsafe-code-guidelines/glossary.html#undefined-behavior) group
+* Unspecified Behavior or Details -- the "miri virtual machine" does not panic, but we don't say what happens. As you point out, there are a number of *reasons* we might not have said what happens yet:
+    * We do not expect to *ever* specify this, because we want the freedom to change it.
+        * Example: Rust struct layout (without a repr attribute) -- not strictly a *behavior*, of course
+        * Example: whether some local variable is materialized on the stack.
+    * We *may* specify this at some point in the future, but there are no plans to do so.
+        * Example: Rust ABI compatibility
+        * Example: What symbols get exported by a DLL
+    * We wish to specify this behavior in the near future, as part of the FFI-unwind project
```
BatmanAoD, line 13: Bullet-point hierarchy doesn't match the heading, "three categories"... I also feel that most of this section feels redundant with the terminology page.

resolved-concerns/README.md
```diff
@@ -0,0 +1,6 @@
+This directory stores "resolved concerns" -- basically, whenever there
+is a non-obvious tradeoff, we try to create a document that describes
+the tradeoff in full, and then documents the resolution we ultimately
+reached. Sometimes these documents include dissents, if not everyone
```
BatmanAoD, line 4: 🚀 🦀

roadmap/c-unwind-abi.md
```diff
@@ -16,30 +16,24 @@ compilers use.
...
-
-[LLVM-UB]: ../spec-terminology.md#LLVM-undefined-behavior-or-LLVM-UB
+roadmap item alone **only** adds the ABI string to Rust. All details
+of how that ABI strings are implemented for various targets are still
+considered [To Be Defined]. As a result, programs that rely on those
+details are not truly stable; you may find that the details change as
+Rust evolves. Eventually, though, we do intend to define many (but not
+all) aspects of how Rust panics and native unwinding interoperate.
+Moreover, we guarantee that unwinding will **not** result in
+[Undefined Behavior] and in particular not [LLVM-UB].
```
BatmanAoD, line 26: This is more of an idea for the future, but maybe we should use a term like "optimizer-UB" or "intermediate-representation-UB" instead of LLVM-UB. The core idea of the term, I think, is that we are avoiding the form of UB that lets compilers make "adversarial" optimizations (in particular, in our case, eliminating landing pads).

spec-terminology.md
```diff
@@ -1,30 +1,88 @@
...
+are considered undefined behavior by LLVM itself. Barring bugs, the
+Rust compiler should never produce LLVM IR that contains LLVM-UB
+unless the behavior in question is *also* UB in Rust. Of course, there
+are bugs in the Rust compiler from time to time, and hence it can
+happen that we generate LLVM IR which contains LLVM-UB even if the
+corresponding Rust source code is meant to be fully defined (see
+e.g. [rust-lang/rust#28728]).  The main reason it is worth separating
+LLVM-UB from the more general form of Rust UB is that, while both
+forms of UB can cause arbitrary things to happen in your
+code. However, as a practical measure, LLVM-UB is much more *likely
```
BatmanAoD, line 40: Looks like these sentences are really supposed to be one sentence that was inadvertently split.

```suggestion
code, LLVM-UB is much more *likely* to do so in practice.
```

(Can't submit a multi-line suggestion yet, but of course this would mean deleting the next line 😄 )

spec-terminology.md
```diff
@@ -1,30 +1,88 @@
...
+[rust-lang/rust#28728]: https://github.com/rust-lang/rust/issues/28728
+
+<a name="unspecified"></a>
+
+## Unspecified behavior
+
+We use the term "unspecified behavior" to refer to behavior that may
+vary across Rust releases, depending on what options are given to the
+compiler, or even -- in extreme cases -- across executions of the Rust
+compiler. However, unlike undefined behavior, the resulting execution
```
BatmanAoD, line 52: It may be worth noting variation across platforms, too.

BatmanAoD (approved):

faq.md
nikomatsakis: I like the idea of using RFCs for these sorts of decisions, as long as we ensure the RFC makes progress. We could draft the RFC in this repo or in the UCG repo -- or in this case maybe just move it directly to the rfcs repo (I'd sort of like to get in the habit of drafting RFCs in the context of a team, though, so maybe draft it here first?).

If you wanted to do that @gnzlbg I'd be supportive!

gnzlbg: So I would prefer to submit a small RFC for the UCGs first, and once that has consensus there, maybe review here that what it guarantees makes sense for these use cases, and only after this has consensus, submit it to the rfc repo.

PR #11, opened 2019-10-16
Draft for the "0th" RFC

nikomatsakis:
This draft includes the basic text for the 0th RFC. It has a few FIXMEs. The summary:

* First, to modify Rust functions with the "C" abi to abort on unwinding, as proposed
  in [rust-lang/rust#52652].
* Second, to introduce the "C unwind" ABI, with virtually all details
  (but not quite all) left as "to be determined".
    * The **intent** of this ABI, however, is to permit Rust panics to
      be "translated" into a "native" unwinding mechanism, though
      precisely which is dependent on the target platform or other
      compiler options.
        * (In practice, this translation is a no-op, since Rust panics
          use the native mechanism, but this is explicitly not
          required for the future.)
    * Folks currently using the "C" ABI to do unwinding can migrate to this new
      "C unwind" ABI, which captures their intentions. Their code is no more stable
      than it was, but they are on the right path.
* Finally, to create a "project group" with the purpose of designing subsequent
  RFCs to flesh out the details of the "C unwind" ABI
    * The "project group" term is newly introduced, but it corresponds
      to a kind of working group -- one with the goal of fleshing out
      a particular proposal or completing a project.
    * We aim to specify how "C unwind" works on major platforms
    * Goal is to enable Rust panics to propagate across native frames
    * And perhaps to enable native exceptions to propagate across Rust frames
    * But not to allow catching or throwing native exceptions from Rust code


rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+themselves do not contain any destructors or landing pads -- they
+simply need to be "skipped over", in effect.
+
+Moreover, Lucet executes only on a narrow range of platforms:
+
+* XXX details
+
+### mozjpeg: XXX
+
+XXX somebody fill this in XXX
```
gnzlbg:
```suggestion
The `mozjpeg-rust` crate provides an idiomatic Rust wrapper over the `libjpeg` library. When this library errors, it calls a user-provided callback that is not allowed to return. The documentation mentions that this callback either aborts or `longjmp` somewhere. The `mozjpeg-rust` crates implements this callback to `panic!` instead, unwinding from the Rust callback into C, and then from C back into Rust.
```

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+using the same "Structured Exception Handling" (SEH) mechanism that is
+used for C++ exceptions and other sorts of errors. As a result, in
+some versions of Rust at least, we found that attempts to longjmp over
+Rust frames triggered unwinding mechanisms. Note that in general a
+longjmp over a Rust frame that contains destructors is [Undefined
+Behavior] -- however, we would like to support longjmp for Rust frames
+that do not contain destructors. At minimum, this group should take
+the behavior of longjmp into account; if it is easy to do, we may also
+introduce mechanisms or RFCs that help to specify the use of longjmp
+in particular cases.
```
gnzlbg, line 109: So this isn't really that related to `longjmp`, although `longjmp` exhibits this issue. As the text says, on these platforms, these longjmps are implemented using "native unwinds". The problem is that on these platforms "native unwinds" can behave differently, depending on, e.g., how they are thrown. 

If this WG attacks the problem of how native unwinds behave in Rust, it would need to consider those potential differences in behavior. 

I'm not sure how to word that into this section which appears more focused on the issue discovered while using rlua on windows.

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+  RFCs to flesh out the details of the "C unwind" ABI
+    * The "project group" term is newly introduced, but it corresponds
+      to a kind of working group -- one with the goal of fleshing out
+      a particular proposal or completing a project.
+    * We aim to specify how "C unwind" works on major platforms
+    * Goal is to enable Rust panics to propagate across native frames
+    * And perhaps to enable native exceptions to propagate across Rust frames
+    * But not to allow catching or throwing native exceptions from Rust code
+
+The first two goals work together. We simul
```
gnzlbg:
```suggestion
The first two goals work together.
```

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+The goal of introducing the "C unwind" ABI is simple: it allows
+existing projects to migrate to the newer ABI and continue to work "as
+well as they ever did". In the meantime, we can tighten up the rules
```
gnzlbg:
```suggestion
well as they ever did" or better, since they will avoid certain optimizations
that are incorrect for their use cases. In the meantime, we can tighten up the rules
```

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+expect the "project group" to largely define its own structure.
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+## The "C" ABI and unwinding
+
+Functions declared with the "C" ABI are not permitted to unwind.  In
+general, any attempt to unwind constitutes [Undefined Behavior], which
+means that the program may do arbitrary things.
```
gnzlbg:
```suggestion
means that the program may do arbitrary things. Rust functions defined
with the "C" ABI are guaranteed to abort if an unwind attempts to
cross the function boundary.
```

gnzlbg: That's mentioned below, but I think it is ok to be more optimistic here and repeat that this RFC removes the amount of UB in Rust by defining some of it to abort.

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+fn foo() {
+    // If `fn_that_is_not_supposed_to_unwind` should actually unwind, 
+    // the result is undefined behavior.
+    fn_that_is_not_supposed_to_unwind();
+}
+```
+
+### Rust functions with C ABI
+
+In the case of Rust functions declared with the "C" ABI, attempts to
```
gnzlbg:
```suggestion
In the case of Rust functions defined with the "C" ABI, attempts to
```

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+}
+```
+
+## The "C unwind" ABI and unwinding
+
+In contrast, functions declared with the "C unwind" ABI **are**
+permitted to unwind. The precise unwinding mechanism in use will
+depend on the target platform, but the intent is to match the
+mechanism used by other systems programming language implementations,
+such as C++.
```
gnzlbg, line 200:
```suggestion
such as C++ or C with `-fexceptions`.
```

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+[summary]: #summary
+
+* First, to modify Rust functions with the "C" abi to abort on unwinding, as proposed
+  in [rust-lang/rust#52652].
+* Second, to introduce the "C unwind" ABI, with virtually all details
+  (but not quite all) left as "to be determined".
+    * The **intent** of this ABI, however, is to permit Rust panics to
+      be "translated" into a "native" unwinding mechanism, though
+      precisely which is dependent on the target platform or other
+      compiler options.
```
gnzlbg, line 16: Which compiler options do you have in mind here ? 

_I think_ that the "native" unwinding mechanism is part of the target ABI, i.e., part of the target triple. So there should be no way to change it via compiler options, doing so requires a completely different target. 

For example, on windows, the mingw targets specify the unwinding ABI as part of their triple in the MinGW C toolchains `...-mingw-seh`, `...-mingw-sjlj`, etc. 

Currently, for Rust, the `x86_64-pc-windows-gnu` and `i686-pc-windows-gnu` use the SEH abi. If we wanted to use the SJLJ abi there, a different target triple, e.g., `x86_64-pc-windows-gnu-sjlj` feels like a better idea, because not only the Rust part of the ABI implementation has to change, but we need to call the C compilers and linkers with different flags, link different platform libraries that use this other ABI, etc. 

There is the potential use case of being able to link both C libraries that use SEH and C libraries that use SJLJ in such a platform, but then "C unwind" isn't enough, because it doesn't let you specify which ABI to use. Instead of maybe trying to make "C unwind" fit this use case, I think it would be better to leave solving this problem to some other language feature, e.g., adding two platform specific ABIs on that target only, like `"C SEH"` and `"C SJLJ"`, or something like that. That kind of ABI information needs to propagate correctly through function pointers and such.

So I would change the text above to:

> The **intent** of this ABI, however, is to permit Rust panics to
> be "translated" into the "native" unwinding mechanism of the platform,
> which is fixed by the platform target specification.

Note that this text is used below as well.

gnzlbg:
One meta-comment is that the RFC suggest that `"C unwind"` is an ABI string that's available on all targets that Rust supports, just like `"system"` or `"C"`, but the RFC does not explicitly state this anywhere, so the reader has to guess. 

If we were to do that, the questions:

* What does `"C unwind"` do on targets that do not support any kind of unwinding (not even in Rust) at all? (AFAICT it would need to be identical to `"C"` ?)
* Will Rust specify / document what `"C unwind"` does for targets for which unwinding is implementable, and Rust does it in a certain way, but the platform doesn't use it nor specify it at all?

I don't think these questions are worth answering, and would prefer the RFC to not leave any room for those questions to appear. Specifying instead that `"C unwind"` is a platform-specific ABI string that's only available on some platforms that can properly support this feature, would mean that these situations would be compilation-errors, and we can re-visit them when a real use case for these appears.

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+- Start Date: 2019-08-29
+- RFC PR: [rust-lang/rfcs#0000](https://github.com/rust-lang/rfcs/pull/0000)
+- Rust Issue: [rust-lang/rust#0000](https://github.com/rust-lang/rust/issues/0000)
+
+# Summary
+[summary]: #summary
+
+* First, to modify Rust functions with the "C" abi to abort on unwinding, as proposed
+  in [rust-lang/rust#52652].
+* Second, to introduce the "C unwind" ABI, with virtually all details
```
gnzlbg, line 11:
```suggestion
* Second, to introduce the target-specific "C unwind" ABI (available on some targets only), with virtually all details
```

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+          required for the future.)
+    * Folks currently using the "C" ABI to do unwinding can migrate to this new
+      "C unwind" ABI, which captures their intentions. Their code is no more stable
+      than it was, but they are on the right path.
+* Finally, to create a "project group" with the purpose of designing subsequent
+  RFCs to flesh out the details of the "C unwind" ABI
+    * The "project group" term is newly introduced, but it corresponds
+      to a kind of working group -- one with the goal of fleshing out
+      a particular proposal or completing a project.
+    * We aim to specify how "C unwind" works on major platforms
```
gnzlbg, line 28:
```suggestion
    * We aim to specify how "C unwind" works on major platforms, adding support for new targets incrementally
```

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+introduce mechanisms or RFCs that help to specify the use of longjmp
+in particular cases.
+
+## Our proposal
+
+This RFC proposes a multi-prong plan to solve this problem:
+
+* First, to modify Rust functions with the "C" abi to abort on unwinding, as proposed
+  in [rust-lang/rust#52652].
+* Second, to introduce the "C unwind" ABI, with virtually all details
```
gnzlbg, line 117:
```suggestion
* Second, to introduce the target-specific "C unwind" ABI (available on some platforms only), with virtually all details
```

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+          required for the future.)
+    * Folks currently using the "C" ABI to do unwinding can migrate to this new
+      "C unwind" ABI, which captures their intentions. Their code is no more stable
+      than it was, but they are on the right path.
+* Finally, to create a "project group" with the purpose of designing subsequent
+  RFCs to flesh out the details of the "C unwind" ABI
+    * The "project group" term is newly introduced, but it corresponds
+      to a kind of working group -- one with the goal of fleshing out
+      a particular proposal or completing a project.
+    * We aim to specify how "C unwind" works on major platforms
```
gnzlbg, line 134:
```suggestion
    * We aim to specify how "C unwind" works on major platforms, adding support for new targets incrementally
```

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+    * Goal is to enable Rust panics to propagate across native frames
+    * And perhaps to enable native exceptions to propagate across Rust frames
+    * But not to allow catching or throwing native exceptions from Rust code
```
gnzlbg, line 31:
```suggestion
    * But we don't know at this point if the working group will end up allowing catching or throwing native exceptions from Rust code
```

@acfoltzer I think something like this is what you meant ?

rfcs/0000-c-unwind-and-project-group-charter.md
```diff
@@ -0,0 +1,496 @@
...
+However, on most platforms, it is expected that the "C" ABI would be
+"compatible" with the "C unwind" ABI. In other words, consider some C
+function `foo` that does not unwind: on most platforms, such a
+function could safely be invoked with either the "C" or the "C unwind"
+ABI. 
+
+However, we do not guarantee such interop: this leaves room for
+platforms where unwinding is transmitted through a special return
+value or other such mechanism. (XXX do any such platforms exist? There
+is the C++ proposal)
```
gnzlbg, line 231: So I'm not sure about these two paragraphs. I suppose that this a defense against "unknown unknowns" but I understand "C unwind" as "the C ABI + unwinding", so I kind of expect it to be equivalent to "C" when `-C panic=abort`, for example. 

The C++ proposal is compatible with the C ABI, and there is also a C proposal for it that adds support for the same C++ feature to C without breaking the current call ABI (I'm not sure if it breaks the unwind ABI on some platforms though, but since the C standard doesn't have a concept of unwinding, it breaks nothing there at least).

BatmanAoD:
Superseded by #28

PR #13, opened 2019-10-17
Langteam sync 10 17

BatmanAoD:


BatmanAoD:
I started with #12 as the basis, so this is actually a much smaller PR than it appears

PR #21, opened 2020-02-15
Blog post announcement of design meeting - draft

BatmanAoD:
[Rendered](https://github.com/BatmanAoD/project-ffi-unwind/blob/BlogPost-announcement/blogposts/inside-rust/01-announcement.md)


blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,162 @@
...
+proposals. Two of these introduce the new `"C unwind"` API, and one does not.
+
+Each proposal specifies the behavior of each type of unwind (Rust `panic!`,
+foreign (e.g. C++), and forced (e.g. `pthread` cancel)) when it encounters an
+ABI boundary under either the `panic=unwind` or `panic=abort` compile-mode.
+
+### 2 APIs, minimal specification
+
+* `"C"` ABI boundary, `panic=<any>`
+  * `panic!`: program aborts
```
Amanieu: Typo? `panic=<any>` causes `panic!` to abort.

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,162 @@
...
+discovered that the question of whether the `"C" ABI should permit unwinding was
+less clear-cut than we had assumed.
+
+If the `"C"` ABI does not permit unwinding, a new ABI, called `"C unwind"`,
+will be introduced specifically to support unwinding.
+
+## Three specific proposals
+
+The project group has narrowed the design space down to three specific
+proposals. Two of these introduce the new `"C unwind"` API, and one does not.
```
Amanieu: "API" should be "ABI". Also a few other places as well.

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,150 @@
...
+  * all other foreign unwinding: undefined behavior
+* `"C unwind"` ABI boundary
+  * With `panic=unwind`: all types of unwinding behave normally
+  * With `panic=abort`: all types of unwinding abort the program
+
+### 2 APIs, forced unwinding always permitted
+
+This is the same as the previous design, except that when compiled with
+`panic=abort`, forced exceptions would *not* be intercepted; that is, the
+program would not abort.
```
Amanieu: This should be clarified: would not be intercepted by the "C unwind" ABI.

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,162 @@
...
+to ensure that Rust implementations may use their own unwinding mechanism,
+which may not be compatible with the platform-provided "native" unwinding
+mechanism. Currently, however, `rustc` uses the native mechanism, and there are
+no plans to change this.
+
+### Forced unwinding
+
+Platform ABIs can define a special kind of unwinding called "forced unwinding."
+This type of unwinding is never supposed to be intercepted by the language
+runtime; so, for instance, C++ destructors should not be invoked.
```
Amanieu: That is not actually true. Destructors are invoked just fine (as intended), but the language is not supposed to catch the forced unwind.

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,162 @@
...
+
+There are two common types of forced unwinding:
+
+* On Windows platforms, `longjmp` is implemented as a forced unwind.
+* On UNIX-like platforms, canceling a `phtread` results in a forced unwind.
```
Amanieu: Typo: pthread. Also I would use `pthread_exit` as an example since that is likely to be more familiar to readers.

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,162 @@
...
+    successfully. This could occur if a panic would've been caught and
+    recovered.
+
+## The primary question: should the `"C"` ABI permit unwinding?
+
+The core question that we would like to decide is whether the `"C"` ABI, as
+defined by Rust, should permit unwinding.
+
+This is not a question we expected to be debating. We've long declared that
+unwinding through Rust's `"C" ABI is undefined behavior. In part, this is
```
Amanieu: Stray backtick `

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,162 @@
...
+because nobody had spent the time to figure out what the correct behavior would
+be, or how to implement it, although (as we'll see shortly) there are other
+good reasons for this choice. 
+
+In any case, in PR #65646, @Amanieu proposed that we could, in fact, simply
+define the behavior of unwinding across `"C"` boundaries. In discussing this,
+discovered that the question of whether the `"C" ABI should permit unwinding was
```
Amanieu: Stray backtick `

blogposts/inside-rust/01-announcement.md
bjorn3: Is it still possible to catch forced unwind, to for example translate to a different unwinding mechanism at the boundary between Rust and C?

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,162 @@
...
+`panic=abort`, forced exceptions would *not* be intercepted; that is, the
+program would not abort.
+
+### 1 API
+
+* `panic=unwind`: unwind behaves normally
+* `panic=abort`:
+  * `panic!`: program aborts
+  * forced unwind, no destructors: unwind behaves normally
+  * all other foreign unwinding: undefined behavior
```
Amanieu: My understanding was that any unwinding, whether foreign or forced, is acceptable as long as there are no destructors. It would be helpful in the table to separate "foreign unwinding, destructors" and "foreign unwinding, no destructors".

blogposts/inside-rust/01-announcement.md
BatmanAoD: Oops. Thank you for the correction, @amanieu.

@bjorn3 Yes, but after the catch block, the unwind will proceed whether or not it is re-thrown.

blogposts/inside-rust/01-announcement.md
BatmanAoD: Should I state the function name, or just say "exiting or canceling a `pthread`? (Does it matter?)

blogposts/inside-rust/01-announcement.md
BatmanAoD: Ah, that may be why Vim highlighting choked. Thanks for catching it.

blogposts/inside-rust/01-announcement.md
bjorn3: In that case I would prefer forced unwind to be UB, or aborting, so that other unwind mechanisms for Rust code are possible. One I once saw which may have an advantage is compiling the DWARF unwind tables to native code performing the unwinding without interpretation.

blogposts/inside-rust/01-announcement.md
BatmanAoD: @Amanieu That wasn't my understanding of the Dropbox Paper doc, but I found it somewhat hard to follow so long after the meeting.

`panic=abort` will not generate landing pads on `"C"` ABI functions. As long as there aren't destructors in those frames, is that sound?

blogposts/inside-rust/01-announcement.md
Amanieu: I would suggest explicitly naming `pthread_exit`.

blogposts/inside-rust/01-announcement.md
Amanieu: The idea with option 3 is that we don't make any distinction between force unwind and other foreign exceptions: they are all treated the same way.

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,162 @@
...
+
+In this table, "UB" stands for "undefined behavior". Some instances of UB can
+be detected at runtime, but the code to do so would impose an undesirable
+code-size penalty; the group recommends generating such code only for debug
+builds. These cases are marked "UB (debug: abort)"
+
+Note that forced unwinding always causes undefined behavior when it traverses
+stack frames with destructors.
+
+|                               | “2 APIs” (Options 1 and 1c), “C”, “panic=unwind” | “2 APIs” (Options 1 and 1c), “C”, “panic=abort” | “2 APIs” (Options 1 and 1c), “C unwind”, “panic=unwind” | “2 APIs, always permit forced” (Option 1), “C unwind”, “panic=abort” | “2 APIs, minimal spec” (Option 1c), “C unwind”, “panic=abort” | “1 API” (Option 3), “panic=unwind” | “1 API” (Option 3), “panic=abort” |
```
Amanieu: Maybe don't refer to "Option N" here. The term isn't used anywhere else.

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,162 @@
...
+| ----------------------------- | ------------------------------------------------ | ----------------------------------------------- | ------------------------------------------------------- | -------------------------------------------------------------------- | ------------------------------------------------------------- | ---------------------------------- | --------------------------------- |
+| panic!                        | abort                                            | abort                                           | unwind                                                  | abort                                                                | abort                                                         | unwind                             | aborts                            |
```
Amanieu: First box should be unwind, not abort.

BatmanAoD: I don't believe so; part of the reason to introduce a second ABI is to _prevent_ unwinds (even `panic!`s) from crossing `extern "C"` boundaries.

blogposts/inside-rust/01-announcement.md
BatmanAoD: @bjorn3 I'm not sure I understand what you're saying. This is the way the platform API _requires_ forced unwinding to work; things like `pthread_exit` and `longjmp`, which we don't want to prohibit, can't behave appropriately without this behavior.

blogposts/inside-rust/01-announcement.md
BatmanAoD: Oops. Copy-and-paste holdover.

blogposts/inside-rust/01-announcement.md
Amanieu: Oh! I got very confused by your use of `panic!`. You're not talking about the behavior of `panic!` itself, you're talking about *an exception generated by `panic!`*. This should be made clearer in the wording.

Also you should use N/A in the table for the panic=abort cases, since `panic!` does not generate an exception in those cases.

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+discussing internally for quite a while. This blog post lays out the arguments
+on each side of the issue and invites the Rust community to join us at the
+[upcoming meeting](meeting-link) to help finalize our decision, which will be
+formalized and published as our first language-change RFC. This RFC will
+propose an "MVP" specification for well-defined cross-language unwinding.
+
+The meeting will be on Monday the 24th at 17:00 UTC, via Zoom:
+
+> Meeting ID: 768 231 760
+> https://mozilla.zoom.us/j/768231760
```
nikomatsakis: For some reason, I feel uncomfortable posting these details here -- can we just link to the [calendar invite](https://calendar.google.com/event?action=TEMPLATE&tmeid=MjlxdW9kdmRjdHJ1cWVtYWlhc2ZvcXIzcGFfMjAyMDAyMjRUMTcwMDAwWiByZWN1ZDRiOW84Y21jMG01cm1yMDMzcDVua0Bn&tmsrc=recud4b9o8cmc0m5rmr033p5nk%40group.calendar.google.com&scp=ALL) instead?

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+  may not necessarily use the system unwinding mechanism
+  * In practice, we do use the system mechanism today, but we would like to
+    reserve the freedom to change this
+* If you enable `-Cpanic=abort`, we are able to optimize the size of binaries
+  to remove most code related to unwinding.
+  * In extreme cases, it should be possible to remove **all** traces of
+    unwinding, when you know that it will not occur.
+  * In practice, most "C" functions are never expected to unwind (because they
+    are written in C, for example, and not in C++).
+    * However, because unwinding is now part of most system ABIs, even C
```
nikomatsakis, line 76: Does anyone know whether this would be a "forced unwind" in practice? i.e., is this the same as `pthread_exit`? (maybe @gnzlbg or @Amanieu knows?)

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+proposals. Two of these introduce the new `"C unwind"` ABI, and one does not.
+
+Each proposal specifies the behavior of each type of unwind (Rust `panic!`,
+foreign (e.g. C++), and forced (e.g. `pthread_exit`)) when it encounters an
+ABI boundary under either the `panic=unwind` or `panic=abort` compile-mode.
+
+### 2 ABIs, minimal specification
+
+* `"C"` ABI boundary, `panic=<any>`
+  * `panic!`: program aborts
```
nikomatsakis: I found this a bit surprising. I think I expected "UB" here. It seems to me that this implies that under `-Cpanic=unwind` we will add some sort of shim around all "C" calls to catch panics and abort, is that correct?

nikomatsakis: Under `-Cpanic=abort`, of course, we can just abort at the point of panic

blogposts/inside-rust/01-announcement.md
nikomatsakis: I think the idea was that we'd behave like `-fno-exceptions`, basically, and hence I think I agree with @Amanieu

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+
+In this table, "UB" stands for "undefined behavior". Some instances of UB can
+be detected at runtime, but the code to do so would impose an undesirable
+code-size penalty; the group recommends generating such code only for debug
+builds. These cases are marked "UB (debug: abort)"
+
+Note that forced unwinding always causes undefined behavior when it traverses
+stack frames with destructors.
+
+|                                                         | “2 ABIs”, “C”, “panic=unwind” | “2 ABIs”, “C”, “panic=abort” | “2 ABIs”, “C unwind”, “panic=unwind” | “2 ABIs, always permit forced”, “C unwind”, “panic=abort” | “2 ABIs, minimal spec”, “C unwind”, “panic=abort” | “1 ABI”, “panic=unwind” | “1 ABI”,  “panic=abort” |
```
nikomatsakis: I think this table would be more readable if we rotated it, so that the *columns* are the kind of unwinding (panic, forced unwind, etc) and the *rows* are the specific scenarios and proposals. It's quite wide as is.

nikomatsakis (commented):
Thanks @BatmanAoD! This is great. My main concern is that the actual explanation around the proposals is pretty light, and I think we've also lost some of the details around the *implications* of the various designs. For example, we don't talk about the fact that 1 ABI means that we sacrifice the "no UB when changing to -Cpanic=abort" constraint -- but what we gain is simplicity. i.e., it's arguably easier to explain our story, even if it is a bit more brittle.

I think these bullets from the end of the paper doc were trying to capture that "core tradeoff":

- Tradeoffs boil down to:
    - If we choose Option 1 or Option 1c:
        - We need to push out this “C” vs “C unwind”
        - We introduce some subtle distinctions:
            - Option 1: forced unwinding vs other exceptions
            - Option 1c: longjmp’ing over “C unwind” doesn’t work
        - We don’t support `pthread_cancel` with destructors in scope
    - If we choose Option 3:
        - We can explain the interactions easily:
            -  `-Cpanic=abort` maps to `-fno-exceptions`
            - otherwise exceptions propagate seamlessly
        - But we risk UB if you combine `-Cpanic=unwind` with `-fexceptions`
            - `-C` options can generally cause UB
            - but panic=unwind/abort is something that can be set in `Cargo.toml` and is particularly “exposed”

blogposts/inside-rust/01-announcement.md
Amanieu: Yes, syscall cancellation due to `pthread_cancel` is basically a forced unwind. It is exactly the same mechanism as the one used by `pthread_exit`.

blogposts/inside-rust/01-announcement.md
Amanieu: Yea I feel this was poorly described (see https://github.com/rust-lang/project-ffi-unwind/pull/21#discussion_r381057844). Basically here we are talking about what happens if an exception generated by `panic!` unwinds to a "C" ABI boundary. This can only happen with `-Cpanic=unwind` since no exception is generated in `-Cpanic=abort`.

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+
+### 2 ABIs, minimal specification
+
+* `"C"` ABI boundary, `panic=<any>`
+  * `panic!`: program aborts
+  * forced unwind, no destructors: unwind behaves normally
+  * all other foreign unwinding: undefined behavior
+* `"C unwind"` ABI boundary
+  * With `panic=unwind`: all types of unwinding behave normally
+  * With `panic=abort`: all types of unwinding abort the program
```
nikomatsakis, line 129: Hmm, actually, I think this entry strays somewhat from what we had in the doc, right? The "minimal spec" there avoided distinguishing forced unwinding from other kinds of exceptions, which simplified life, but it meant that we couldn't insert helpful shims to catch unwinding and abort, because we couldn't locally tell when unwinding might occur. 

In any case, this setup as described here means that one cannot `longjmp` over a "C unwind" boundary in panic=abort mode, which seems not great. This is why I think we carved out a special case in the paper doc which made "forced unwinding with destructors" UB *even in panic=unwind*.

Amanieu: This is option 1c described in the doc. Option 1 is the one below which carves out the special exception for forced unwinding.

blogposts/inside-rust/01-announcement.md
BatmanAoD: That's an excellent point. Yes, the rows are different types of unwinding, so the `panic!` row is indeed for the unwinding operation, not the behavior of `panic!()`.

For `panic=abort`, it seems immaterial to me from a user perspective that the abort occurs before the `panic!()` macro actually starts unwinding the stack; in either case, the behavior is well defined and the ultimate outcome is that the program aborts. But you're right that the difference between "abort and function boundary" and "abort without unwinding" is observable in practice depending on what destructors are executed before the unwind operation reaches (or would reach) an `extern "C"` API boundary.

So perhaps the entries for that column should say something like "aborts without any unwinding"?

blogposts/inside-rust/01-announcement.md
BatmanAoD: I can take these out. I already included a link to an "Are We Meeting Yet" page, which has the advantage of not requiring a Google log-in to view. Should I keep that link, or replace it with the invite link?

blogposts/inside-rust/01-announcement.md
BatmanAoD: Yes, the behavior as I've written it up here would require such shims. I may have misunderstood the proposal from the Paper Doc; I do remember discussing avoiding such shims if possible.

blogposts/inside-rust/01-announcement.md
```diff
@@ -0,0 +1,166 @@
...
+Platform ABIs can define a special kind of unwinding called "forced unwinding."
+This type of unwinding operation cannot be stopped prematurely using language
+mechanisms such as Rust's `catch_unwind` or C++'s `catch`. (These mechanisms
+can intercept forced unwinds, but the runtime will resume unwinding regardless
+of the content of the intercepting code, unless the program is aborted).
+
+There are two common types of forced unwinding:
+
+* On Windows platforms, `longjmp` is implemented as a forced unwind.
+* On UNIX-like platforms, `pthread_exit` results in a forced unwind.
```
acfoltzer: IIRC
```suggestion
* On UNIX-like platforms, `pthread_exit` and `pthread_cancel` result in a forced unwind.
```

blogposts/inside-rust/01-announcement.md
BatmanAoD: Fixed; discussion here: https://rust-lang.zulipchat.com/#narrow/stream/210922-project-ffi-unwind/topic/weekly.20meeting/near/188664814

blogposts/inside-rust/01-announcement.md
BatmanAoD: Done! Good suggestion.

blogposts/inside-rust/01-announcement.md
BatmanAoD: @nikomatsakis I'm about to merge this as-is, but Amanieu will make a PR with further changes, so once that's open you can weigh in again on which link(s) to include (Google and/or arewemeetingyet).

blogposts/inside-rust/01-announcement.md
BatmanAoD: @bjorn3 I'm merging this draft as-is, but there will be a second PR with changes before the blog post is finalized, and then of course there will be the meeting on Monday, so you'll have a few more opportunities to chime in!

blogposts/inside-rust/01-announcement.md
BatmanAoD: I've reworded; it should be clearer.

PR #22, opened 2020-02-21
More fixups to blog post

Amanieu:
Some edits on top of #21 

cc @BatmanAoD @nikomatsakis @acfoltzer


blogposts/inside-rust/01-announcement.md
```diff
@@ -53,8 +53,8 @@ no plans to change this.
 Platform ABIs can define a special kind of unwinding called "forced unwinding."
 This type of unwinding operation cannot be stopped prematurely using language
 mechanisms such as Rust's `catch_unwind` or C++'s `catch`. (These mechanisms
-can intercept forced unwinds, but the runtime will resume unwinding regardless
-of the content of the intercepting code, unless the program is aborted).
+can sometimes intercept forced unwinds, but they must resume the unwind
+ otherwise the runtime will abort the program).
```
BatmanAoD, line 57: Looking at the spec, it seems there is no suggestion of automatically aborting:

> A catch-all block may be executed during forced unwinding. For instance, a `longjmp may execute code in a catch(...) during stack unwinding. However, if this happens, unwinding will proceed at the end of the catch-all block, whether or not there is an explicit rethrow. 

....but I suppose a compiler could automatically insert some logic to abort rather than letting the unwind be resumed.

I'm trying to determine whether GCC (and by extension, maybe C++ in general?) have the behavior you're describing, where a non-rethrowing `catch` block would `abort` instead of automatically resuming the unwind. But on my Mac, it seems like I don't see destructors being run whether or not I rethrow the `pthread_exit` exception. In fact, I don't even see a print statement in the `catch` block itself. http://coliru.stacked-crooked.com/a/ee07fca96fe69c47

blogposts/inside-rust/01-announcement.md
```diff
@@ -86,6 +86,8 @@ There are two common types of forced unwinding:
   * It may, of course, cause programs to abort that used to execute
     successfully. This could occur if a panic would've been caught and
     recovered.
+* Changing the ABI (the `"C"` in `extern "C"`) of functions in the libc crate is
+  a breaking change: function pointers of different ABIs have different types.
```
BatmanAoD: We consider this to be an unacceptable cost, correct? (This should probably be explicit either way.)

blogposts/inside-rust/01-announcement.md
```diff
@@ -128,31 +130,43 @@ Throughout, the  unwind generated by `panic!` will be referred to as
...
+### Proposal 3: No new ABI
 
 * `panic=unwind`: unwind behaves normally
 * `panic=abort`:
   * `panic`-unwind: does not exist; `panic!` aborts the program
   * foreign unwinding, no destructors: unwind behaves normally
   * foreign unwinding with destructors: undefined behavior
 
+The main advantage of this proposal is its simplicity: there is only one ABI and
+the behavior of `panic=unwind` is identical to that of `-fno-exceptions` in C++.
```
BatmanAoD:
```suggestion
the behavior of `panic=abort` is identical to that of `-fno-exceptions` in C++.
```

blogposts/inside-rust/01-announcement.md
```diff
@@ -128,31 +130,43 @@ Throughout, the  unwind generated by `panic!` will be referred to as
...
 ## Comparison table for the proposed designs
 
 In this table, "UB" stands for "undefined behavior". Some instances of UB can
 be detected at runtime, but the code to do so would impose an undesirable
 code-size penalty; the group recommends generating such code only for debug
 builds. These cases are marked "UB (debug: abort)"
 
-Note that forced unwinding always causes undefined behavior when it traverses
-stack frames with destructors.
+Note that unwinding through a frame without running its destructors (e.g. because they have been optimized out by `panic=abort`) is always undefined behavior, except if the frame has no destructors.
```
BatmanAoD:
```suggestion
Note that unwinding through a frame that has destructors without running those destructors (e.g. because they have been optimized out by `panic=abort`) is always undefined behavior.
```

blogposts/inside-rust/01-announcement.md
Amanieu: Basically only glibc on Linux uses forced unwinding for pthread_exit. Even musl on Linux just exits the thread directly without unwinding.

The abort happens because glibc defines the destructor of the exception object passed to _Unwind_ForcedUnwind to abort if the exception object is ever destroyed (which happens if you catch it without rethrowing it).

blogposts/inside-rust/01-announcement.md
Amanieu: I think so. Unless we want to play around with fancy coercion rules... (but I'd rather not)

blogposts/inside-rust/01-announcement.md
BatmanAoD: Ah, that makes a lot of sense. It seems like it's probably worth noting that the statement about `pthread_exit` only applies to glibc on Linux, though.

blogposts/inside-rust/01-announcement.md
BatmanAoD: Yeah, then let's just say something like:

```
  * We cannot change the ABI (the `"C"` in `extern "C"`) of functions in the libc
    crate, because this would be a breaking change: function pointers of different
    ABIs have different types.
```

BatmanAoD:
I think this looks good. I'm going to merge for now, then edit it again to mark the date as TBD.

PR #24, opened 2020-02-23
Add 'advantage' paragraph for proposal 2

BatmanAoD:


Amanieu:
Add a small sentence about the downside: The downside of this proposal is that the FFI unwinding rules become more complicated since they need to distinguish between different types of foreign exceptions.

PR #28, opened 2020-04-02
"C unwind" RFC

BatmanAoD:
Based on "option 2" from the Internals post


rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,233 @@
...
+
+Introduces a new ABI string, `"C unwind"`, to enable unwinding from other
+languages (such as C++) into Rust frames and from Rust into other languages.
+
+Additionally, defines the behavior for a limited number of previously-undefined
+cases when an unwind operation reaches a Rust function boundary with a
+non-`"Rust"`, non-`"C"` ABI.
+
+This RFC does not define the behavior of `catch_unwind` in a Rust frame being
+unwound by a foreign exception.
```
Amanieu: Actually the decision was to not specify the behavior of foreign exceptions unwinding through any Rust frame, whether they include `catch_unwind` or not.

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,233 @@
...
+Rust frames; on Windows, `longjmp` is implemented via unwinding.
+
+<!-- TODO: links to prior discussions? -->
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When declaring an external function that may unwind, such as an entrypoint to a
+C++ library or a `libc` function that may invoke `pthread_exit`, use
+`"C unwind"`:
```
Amanieu: Forced unwinding should just use plain `"C"`. Also the current plan is to only specify the behavior of Rust exceptions unwinding through foreign code. C++ exceptions unwinding through Rust code is still UB.

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,233 @@
...
+cross into Rust via an `extern "C unwind"` function declaration, safely unwind
+the Rust frames, and cross back into C++ (where it may be caught) via a Rust
+`"C unwind"` function definition.
+
+Conversely, languages that support the native unwinding mechanism, such as C++,
+may be "sandwiched" between Rust frames, so that Rust `panic`s may safely
+unwind the C++ frames, if the Rust code declares both the C++ entrypoint and
+the Rust entrypoint using `"C unwind"`.
+
+## Forced unwinding
```
Amanieu, line 129: The behavior of forced unwinding is orthogonal to `"C"` vs `"C unwind"`: it works for both as long as you don't have destructors. As such I feel that it should be left to a separate RFC, or to its own section in this RFC.

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,233 @@
...
+
+Prior to this RFC, any unwinding operation that crossed an `extern "C"`
+boundary, either from a `panic!` "escaping" from a Rust function defined with
+`extern "C"` or by entering Rust from another language via an entrypoint
+declared with `extern "C"`, caused undefined behavior.
+
+This RFC retains most of that undefined behavior, with two exceptions:
+
+* With the `panic=abort` runtime, `panic!` will cause an `abort` if it would
+  otherwise "escape" from a function defined with `extern "C"`.
```
Amanieu, line 169: No, that behavior is only with `panic=unwind`. We don't want Rust panics to escape into `extern "C"` since that would be UB.

rfcs/0000-c-unwind-abi.md
BatmanAoD: That's... surprising. Why is that?

rfcs/0000-c-unwind-abi.md
BatmanAoD: Oops, yes, of course. Good catch.

rfcs/0000-c-unwind-abi.md
BatmanAoD: I think something I've been overlooking (because we didn't represent it this way in the behavior table) is that forced unwinding over destructors is UB whether or not `"C unwind"` is used. Is that correct?

If not, then it's not truly orthogonal, I think.

rfcs/0000-c-unwind-abi.md
BatmanAoD: Per discussion in Zulip, we'll go ahead and make the case without `catch_unwind` well-defined.

rfcs/0000-c-unwind-abi.md
BatmanAoD: I believe I've fixed this.

rfcs/0000-c-unwind-abi.md
BatmanAoD: I've gone through the document and changed most of the text related to forced unwinding. I think it's (more) correct now.

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,273 @@
...
+  destructors (i.e. `Drop` types) are unwound.
+
+## Interaction with `panic=abort`
+
+When compiling with the `panic=abort` panic runtime, there are some types of
+unwinding that are not guaranteed cause the program to abort:
+
+* Forced unwinding: Rust provides no mechanism to catch this type of unwinding.
+* Unwinding from another language into Rust if the entrypoint to that language
+  is declared with `extern "C"` (contrary to the guidelines above)
```
Amanieu: It should be clarified that this is UB.

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,273 @@
...
+
+* With the `panic=unwind` runtime, `panic!` will cause an `abort` if it would
+  otherwise "escape" from a function defined with `extern "C"`.
+* Forced unwinding is safe with `extern "C"` as long as no frames with
+  destructors (i.e. `Drop` types) are unwound.
+
+## Interaction with `panic=abort`
+
+When compiling with the `panic=abort` panic runtime, there are some types of
+unwinding that are not guaranteed cause the program to abort:
```
Amanieu: You accidentally the English.

rfcs/0000-c-unwind-abi.md
BatmanAoD: Reworded/reorganized this section

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,273 @@
...
+[forced-unwinding]: #forced-unwinding
+
+This is a special kind of unwinding used to implement `longjmp` on Windows and
+`pthread_exit` in `glibc`. A brief explanation is provided in [this Inside Rust
+blog post][inside-rust-forced]. This RFC distinguishes forced unwinding from
+other types of foreign unwinding.
+
+Since language features implemented using forced unwinding on some platforms
+use other mechanisms on other platforms, Rust code should not rely on forced
+unwinding to invoke destructors (calling `drop` on `Drop` types). External
```
bjorn3: This seems to imply that forced unwinding accross frames with `Drop` types is allowed, rather than being UB as said below.

BatmanAoD: When destructors don't run, that's undefined behavior. But in the case where `"C unwind"` is used, forced unwinding across destructors will in fact invoke destructors under this proposal, so it's _technically_ not UB according to this proposal.

@Amanieu @nikomatsakis @joshtriplett This seems like an oddity to me; the existence, or not, of UB should generally be determined by the language virtual machine model, not by the platform. So even though we expect this case to work "in practice", it seems that the language spec should leave it UB; do you agree? (This would also resolve the concern that the proposal has a case where changing from `panic=unwind` to `panic=abort` introduces UB.)

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,273 @@
...
+There are some Rust projects that need cross-language unwinding to provide
+their desired functionality. One major example is WASM interpreters, including
+the Lucet and Wasmer projects.
+
+There are also existing Rust crates (notably, wrappers around the `libpng` and
+`libjpeg` C libraries) that rely on compatibility with the native exception
+handling mechanism in GCC, LLVM, and MSVC.
+
+Additionally, there are libraries such as `rlua` that rely on `longjmp` across
+Rust frames; on Windows, `longjmp` is implemented via unwinding.
```
nikomatsakis: This is mildly misleading, right? In particular, you should be able to use longjmp portably with the standard "C" ABI, as long as you have no destructors? (I realize this is only the motivation, but ..)

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,273 @@
...
+# Motivation
+[motivation]: #motivation
+
+There are some Rust projects that need cross-language unwinding to provide
+their desired functionality. One major example is WASM interpreters, including
+the Lucet and Wasmer projects.
+
+There are also existing Rust crates (notably, wrappers around the `libpng` and
+`libjpeg` C libraries) that rely on compatibility with the native exception
+handling mechanism in GCC, LLVM, and MSVC.
```
nikomatsakis: Can we be a bit more specific? IIRC, these crates are using the ability to "panic across C frames"? (because there is no other way to abort processing), is that right?

That *is* a case that would require "C unwind" ABI

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,273 @@
...
+
+Additionally, there are libraries such as `rlua` that rely on `longjmp` across
+Rust frames; on Windows, `longjmp` is implemented via unwinding.
+
+The desire for this feature has been previously discussed on other RFCs,
+including #2699 and #2753.
```
nikomatsakis:
```suggestion
including [#2699](https://github.com/rust-lang/rfcs/pull/2699) and [#2753](https://github.com/rust-lang/rfcs/pull/2573).
```

links ftw

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,273 @@
...
+
+# Guide-level explanation
```
nikomatsakis, line 88: I think I would move some of the "older behavior" material into the Motivation

rfcs/0000-c-unwind-abi.md
nikomatsakis: There is some trickiness here based on whether we define the behavior in terms of *forced exceptions* or in terms of *particular, well-known C functions*.

In practice, it seems like forced exceptions seem to be "highly correlated" with abrupt jumps, like longjmp and pthread-exit, that are also implemented in different ways across platforms. Let's them "abrupt-jump" operations.

I think the core invariant is:

* Let's define a Rust stack frame as "inert" if it it has no pending destructors and does not correspond to the `catch_panic` intrinsic. 
* It is UB to unwind or otherwise deallocate a Rust stack frame without running its destructors.
* Forced unwind across a "C" ABI boundary is:
    * not a problem if all Rust stack frames are inert
    * UB otherwise (because it would cause Rust destructors not to execute)
* Forced unwind across a "C unwind" ABI boundary is:
    * OK if all stack frames are "inert"
    * "Not Yet Specified" otherwise, because we are still working on what it means for a foreign exception to propagate across non-inert Rust frames

We could then say that the implementation of operations like longjmp and pthread-exit varies by platform. The only **portable way** to use those functions that is not UB is 

* with the "C" or "C unwind" ABIs and to only drop inert Rust frames

In particular, on non-Windows platforms, longjmp will immediately deallocate frames, which is UB unless those frames are inert. On Windows, longjmp will trigger a forced exception, and the interaction of foreign exceptions with non-inert frames is Not Yet Specified.

Note 1: recall that "Not Yet Specified" just means UB that we maybe intend to change to something defined later (but it's UB until we do so).

Note 2: I believe that C++ just kind of jumps to the final point, and directly defines longjmp across C++ frames as UB if those frames are "pending" (per my definition above). I think it's ok to define the core operations and say that the behavior of longjmp etc ultimately depends on what core operations are used to implement it.

Thoughts?

nikomatsakis (commented):
Gave a quick read. Nice job, thanks for whipping this up. Here are some initial thoughts.

rfcs/0000-c-unwind-abi.md
BatmanAoD: Hm, I thought we had agreed that there's no reason to leave the foreign-exceptions case not-yet-specified. The specification (in this RFC) seems straightforward to me; if there is a remaining doubt about the specifics of the specification, then I'd be okay with leaving it not-yet-specified, but I don't know that that's the case.

Do you want to include new definitions for destructor-free, catch-free frames and for frames that don't fit those criteria in the RFC itself? If so, I'm not quite sold on the "inert"/"pending" terminology. Those words aren't the inverse of each other in common language, and "inert" doesn't strike me as an obvious analogy. I think perhaps "disposable" might be better, since the frames can be "disposed" without side-effects. Or, we can simply apply Rust's trait name to the frames themselves: frames with `Drop` locals are "drop" frames. (The nonstandard verb-as-adjective syntax is intentional, to signal that it's a technical term.)

A Rust compiler (maybe not `rustc` specifically) could even emit code to avoid invoking destructors in the case of forced-unwind; that would make thread-termination and `longjmp` have "real" (unpredictable/dangerous in practice) UB across such "pending"/"non-disposable" frames on all platforms. Should the RFC specifications (or the reference, or a future language spec) prohibit this? I'm not sure there would be a benefit to doing so.

This leaves me with the opinion that we should say "forced unwinding across drop frames is always UB." That way, we can make the equivalent statement that "pthread-cancelling and `longjmp` are UB when they cross drop frames, on all platforms."

BatmanAoD: I've revised the RFC to this effect. I think it simplifies the specification a fair amount, actually.

rfcs/0000-c-unwind-abi.md
BatmanAoD: I think that's still a motivation for specifying that it's safe, right? It is in fact already safe even with the current implementation, but I don't think that has been specified in any other RFC or in the Reference. I can add a sentence to the effect that this particular use-case is already *de facto* supported.

rfcs/0000-c-unwind-abi.md
BatmanAoD: I think I've clarified this.

rfcs/0000-c-unwind-abi.md
BatmanAoD: Hm, I thought that wasn't necessary since this will be posted in the RFC repository, so the links will work automatically when rendered by GitHub in that repo. I suppose it's better to be explicit, though, especially since the current version is in this repo.

rfcs/0000-c-unwind-abi.md
BatmanAoD: @nikomatsakis marking this resolved for now; please un-resolve if you think this is the wrong decision. (@Amanieu you may also disagree with my decision.)

rfcs/0000-c-unwind-abi.md
BatmanAoD: Do you mean the paragraph at the beginning of the "Changes to `extern "C"` behavior" section?

bjorn3:
Can the personality function detect if a function requires cleanup? In that case it would be nice to abort the process on forced-unwinding of drop frames on at least some systems.

BatmanAoD:
@bjorn3 Amanieu might be able to answer that, but even if it can't, the destructors themselves could do that. It's sort of expensive to insert that logic everywhere, though, so we'd only want to do that in debug mode.

rfcs/0000-c-unwind-abi.md
nikomatsakis: not sure what I meant anymore, I'll recreate the comment if I have "the feeling" again :)

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,288 @@
...
+The desire for this feature has been previously discussed on other RFCs,
+including [#2699][rfc-2699] and [#2753][rfc-2753].
+
+[rfc-2699]: https://github.com/rust-lang/rfcs/pull/2699
+[rfc-2753]: https://github.com/rust-lang/rfcs/pull/2573
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+When declaring an external function that may unwind, such as an entrypoint to a
```
nikomatsakis, line 97: I would probably kick this off like so:

When one declares a `extern "C"` function, the general rule is that this function may not unwind (although there are some very narrow exceptions, as we'll see). This means, for example, that the function cannot throw an uncaught C++ exception, and it also cannot invoke Rust code that may panic (unless that panic is caught with `catch_panic`).

If you wish to invoke an external function that may unwind, use the "C unwind" ABI instead, like so:

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,288 @@
...
+  // ...
+```
+
+Using the `"C unwind"` ABI to "sandwich" Rust frames between frames from
+another language (such as C++) allows an exception initiated in a callee frame
+in the other language to traverse the intermediate Rust frames before being
+caught in the caller frames. I.e., a C++ exception may be thrown,
+cross into Rust via an `extern "C unwind"` function declaration, safely unwind
+the Rust frames, and cross back into C++ (where it may be caught) via a Rust
+`"C unwind"` function definition.
```
nikomatsakis, line 122: I think that, at present, we were saying that it was still UB for a C++ exception to propagate across Rust frames, right?

Or, at minimum, we've not defined what the interaction with `catch_unwind` should be.

Though I think I would be ok with saying "catch-unwind will abort" and then discussing the possibility of adding some new mechanism later.

This would basically mean that one can "propagate C++ exceptions" in the manner described here, but only if you carefully control the Rust code in between and you know that it will not attempt to use `catch_unwind`.

nikomatsakis: I'd be curious to hear what @rust-lang/lang thinks about this, as well as @Amanieu -- how far do we want to go in terms of defining the interaction between "foreign" unwinding like C++ and Rust frames?

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,288 @@
...
+
+Conversely, languages that support the native unwinding mechanism, such as C++,
+may be "sandwiched" between Rust frames, so that Rust `panic`s may safely
+unwind the C++ frames, if the Rust code declares both the C++ entrypoint and
+the Rust entrypoint using `"C unwind"`.
+
```
nikomatsakis, line 128: I think it might be helpful to have a list of scenarios and a description of what works and what doesn't.

| Scenario | Works |
| --- | --- |
| C++ exception propagating across Rust frames | ✔️ |
| C++ exception propagating across `catch_unwind` | ❌ -- Currently UB, TBD |
| Rust panic propagating across a "C unwind" frame | ✔️ |

the challenge here though is how to really do this table, the textual descriptions aren't that great. Also I think it'd be worth being specific about showing the ABI details.

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,288 @@
...
+Since language features and library functions implemented using forced
+unwinding on some platforms use other mechanisms on other platforms, Rust code
+cannot rely on forced unwinding to invoke destructors (calling `drop` on `Drop`
+types). In other words, a forced unwind operation on one platform will simply
+deallocate Rust frames without true unwinding on other platforms.
+
+This RFC specifies that it is undefined behavior to cross Rust frames with
+destructors via such language features and library functions, regardless of the
+platform.
```
nikomatsakis: This seems good to me. It's probably worth stating explicitly that this is true for both "C" and "C unwind" functions -- and also that it IS OK to unwind if there are no destructors.

Maybe something like:

"Forced unwinding" works differently than ordinary unwinding. For both "C" and "C unwind" functions, unwinding a Rust frame with forced unwinding is:

* *undefined behavior* if there are destructors in scope (or a `catch_unwind` call)
* *defined behavior* otherwise

As an example, this means that Rust code can invoke `longjmp` (or `C` functions which invoke `longjmp`) using the "C" ABI, and that longjmp can unwind Rust frames, so long as those frames do not contain any pending destructors or make use of `catch_unwind`. If those Rust frames *do* contain pending destructors, then invoking `longjmp` would be *undefined behavior* (and hence a bug).

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,288 @@
...
+frames with C functions that deallocate frames (i.e. functions that may use
+forced unwinding on specific platforms) is specified as follows:
+
+* **When deallocating Rust frames without destructors:** frames are safely
+  deallocated; no undefined behavior
+* **When deallocating Rust frames with destructors:** undefined behavior
+
+No subtype relationship is defined between functions or function pointers using
+different ABIs. This RFC also does not define coercions between `"C"` and
+`"C unwind"`.
```
nikomatsakis, line 215: Maybe we should include a coercion from "C" to "C unwind"

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,288 @@
...
+
+Our reasons for preferring the current proposal are:
+
+* Introducing a new ABI makes reliance on cross-language exception handling
+  more explicit.
+* `panic=abort` can be safely used with `extern "C unwind"` (there is no
+  undefined behavior except with improperly used forced unwinding), but `extern
+  "C"` has more optimization potential (eliding landing pads). Having two ABIs
+  puts this choice in the hands of users.
+  * Additionally, there are very few cases in which the current proposal
```
nikomatsakis: "very few"? Are there any?

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,288 @@
...
+gives `longjmp` and `pthread_exit` surprising behavior on some platforms, but
+avoids that cause of undefined behavior in the current proposal.
+
+The other proposal in the blog post, "option 3", is dramatically different. In
+that proposal, foreign exceptions are permitted to cross `extern "C"`
+boundaries, and no new ABI is introduced.
+
+[inside-rust-proposals]: https://blog.rust-lang.org/inside-rust/2020/02/27/ffi-unwind-design-meeting.html#three-specific-proposals
+
+## Reasons for the current proposal
```
nikomatsakis, line 265: I was going to say that we should list out the design constraints that led us here. I guess that this section goes in that direction. This was my list:

* Able to fully remove all unwinding machinery from panic=abort builds. 
* Able to change Rust panics to an alternative implementation beyond native system unwinding if desired, by inserting "translation shims" at "C unwind" call sites. This is possible under either implementation, but such shims would be much more widespread with a single ABI.
* Repair the soundness hole when defining safe Rust functions with "C" ABI that panic (by aborting or translating the panic into a foreign exception). All designs did this equally well, either by aborting or by simply propagating the panic naturally.
* Permit a Rust panic to propagate across foreign frames and re-enter Rust code.
* Support longjmp/forced exceptions with "C" ABI, as many common C functions may unwind in this way. This proposal handles this case so long as there are no destructors, which was deemed adequate, as these features are highly platform dependent and not really suitable for widespread use.

rfcs/0000-c-unwind-abi.md
nikomatsakis: I guess there is a similar table in the "detailed design" section.

nikomatsakis: I feel like I personally would like to see a list of scenarios we considered and which ones work and which ones don't. At least, as a user I would find that very useful.

nikomatsakis: To some extent I am doing the work I think we'd need to do to document this properly in the reference or other docs, but I think that is indeed the role of this section.

rfcs/0000-c-unwind-abi.md
BatmanAoD: Adopted with a bit of rephrasing.

rfcs/0000-c-unwind-abi.md
BatmanAoD: In the interest of simplicity for the initial specification, I would prefer to leave the interaction with `catch_unwind` TBD (formally UB) for now, even if the initial implementation `abort`s.

rfcs/0000-c-unwind-abi.md
BatmanAoD: Adopted with some rephrasing.

rfcs/0000-c-unwind-abi.md
BatmanAoD: I've added some verbiage about our usage of "TBD".

rfcs/0000-c-unwind-abi.md
BatmanAoD: Hm -- for simplicity, for now, I'd rather not. I could probably be persuaded otherwise, but I think we'd have a better idea of whether this would actually be substantially beneficial after `"C unwind"` has been available to the general public for a while.

rfcs/0000-c-unwind-abi.md
BatmanAoD: Oops. There were until this weekend. Thanks for catching that.

rfcs/0000-c-unwind-abi.md
BatmanAoD: Hm. I'm not sure I see a significant difference between what you're asking for and what's provided in the current draft. Though I do think that green/red checkmarks in the tables would be pretty... 😄

rfcs/0000-c-unwind-abi.md
BatmanAoD: Hm, several of these are listed in the blog post; perhaps we should link to that? In this section, I focused on differences compared to the other options we discussed in the design-meeting, rather than on our non-negotiable constraints.

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,347 @@
...
+This RFC does not define the behavior of `catch_unwind` in a Rust frame being
+unwound by a foreign exception. This is something the [project
+group][project-group] would like to specify in a future RFC; as such, it is
+"TBD" (see ["Unresolved questions"][unresolved-questions]).
+
+# Motivation
+[motivation]: #motivation
+
+There are some Rust projects that need cross-language unwinding to provide
+their desired functionality. One major example is WASM interpreters, including
```
acfoltzer:
```suggestion
their desired functionality. One major example is Wasm interpreters, including
```

rfcs/0000-c-unwind-abi.md
```diff
@@ -0,0 +1,347 @@
...
+including [#2699][rfc-2699] and [#2753][rfc-2753].
+
+# Guide-level explanation
+[guide-level-explanation]: #guide-level-explanation
+
+Functions declared with `extern "C"` are generally not permitted to unwind
+(with some very narrow exceptions, described [below][forced-unwinding]).
+This means, for example, that such a function cannot throw an uncaught C++
+exception, and it also cannot invoke Rust code that may panic (unless that
+panic is caught with `catch_panic`).
```
acfoltzer:
```suggestion
panic is caught with `catch_unwind`).
```

acfoltzer (commented):
Thank you so much for writing this up. This looks great to me, and would address all of our needs with Lucet.

rfcs/0000-c-unwind-abi.md
nikomatsakis: We could link to the blog post, although I think it's good for the RFC to be reasonably self-contained. Or we could just copy some of that material into a different section. Honestly, I think I would typically put these sorts of constraints into the introduction

rfcs/0000-c-unwind-abi.md
nikomatsakis: well, I don't think the draft actually lists out scenarios very thoroughly, and arranging them a table makes them easier to understand. It's more a matter of degree than anything.

rfcs/0000-c-unwind-abi.md
nikomatsakis: So do you think we should say:

* propagating a C++ exception is OK (with "C unwind" and "panic=unwind")
* but not if there is `catch_unwind`?

This does imply that panic=abort would have to intercept the C++ exception and abort.

BatmanAoD: Yes, I think that's accurate. I think we already had the `panic=abort` requirement?

BatmanAoD:
Merging, though this is not _quite_ ready to be submitted to the `rfcs` repo.

PR #29, opened 2020-05-30
POF terminology, links & summary for older discussions

BatmanAoD:
This should wrap up all the TODOs that were holding us back from opening up an RFC repo PR.

[Rendered](https://github.com/BatmanAoD/project-ffi-unwind/blob/Rfc-POF-terminology/rfcs/0000-c-unwind-abi.md#older-discussions-about-unwinding-through-extern-c-boundaries)

### DIFF
--- a/rfcs/0000-c-unwind-abi.md
+++ b/rfcs/0000-c-unwind-abi.md
@@ -16,6 +16,10 @@ Additionally, we define the behavior for a limited number of
 previously-undefined cases when an unwind operation reaches a Rust function
 boundary with a non-`"Rust"`, non-`"C unwind"` ABI.
 
+As part of this specification, we introduce the term ["Plain Old Frame"
+(POF)][POF-definition]. These are frames that may be safely deallocated with
+`longjmp`.
+
 This RFC does not define the behavior of `catch_unwind` in a Rust frame being
 unwound by a foreign exception. This is something the [project
 group][project-group] would like to specify in a future RFC; as such, it is
@@ -36,8 +40,8 @@ native exception mechanisms in GCC, LLVM, and MSVC.
 Additionally, there are libraries such as `rlua` that rely on `longjmp` across
 Rust frames; on Windows, `longjmp` is implemented via [forced
 unwinding][forced-unwinding]. The current `rustc` implementation makes it safe
-to `longjmp` across Rust frames without `Drop` types, but this is not formally
-specified in an RFC or by the Reference.
+to `longjmp` across Rust [POFs][POF-definition] (frames without `Drop` types),
+but this is not formally specified in an RFC or by the Reference.
 
 The desire for this feature has been previously discussed on other RFCs,
 including [#2699][rfc-2699] and [#2753][rfc-2753].
@@ -74,8 +78,8 @@ how well the current design satisfies these constraints.
 * **Enable error handling with `longjmp`:** As mentioned above, some existing
   Rust libraries use `longjmp`. Despite the fact that `longjmp` on Windows is
   [technically a form of unwinding][forced-unwinding], using `longjmp` across
-  Rust frames [is safe][longjmp-pr] with the current implementation of `rustc`,
-  and we want to specify that this will remain safe.
+  Rust [POFs][POF-definition] [is safe][longjmp-pr] with the current
+  implementation of `rustc`, and we want to specify that this will remain safe.
 * **Do not change the ABI of functions in the `libc` crate:** Some `libc`
   functions may invoke `pthread_exit`, which uses [a form of
   unwinding][forced-unwinding] in the GNU libc implementation. Such functions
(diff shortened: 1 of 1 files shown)


mjbshaw:
After reading about POFs, I have a question: why is unwinding through a non-POF (that only contains `Drop`s, not `catch_unwind`s) UB? I'm not necessarily trying to argue that it shouldn't be UB; rather I'm suggesting that the RFC explain *why* unwinding through a non-POF is UB. When I first read the changes here I had the thought "Unwinding on a non-POF is kinda like `core::mem::forget`ing everything, and [the docs](https://doc.rust-lang.org/core/mem/fn.forget.html) for that say 'Rust's safety guarantees do not include a guarantee that destructors will always run', so why is unwinding through a non-POF UB?"

I've given this a bit of thought and I think I was able to answer my own question, but I think a few sentences in the RFC that give a reason or two for *why* unwinding through a non-POF is UB would be helpful. Calling out why this is different from `core::mem::forget` would be nice.

BatmanAoD:
@mjbshaw Only _forced_ unwinding, which is the mechanism that implements `longjmp` on Windows platforms, causes UB when applied to non-POFs. Is that what you meant, or is the draft verbiage somehow unclear?

It may be reasonable to explain why it's necessarily UB for a forced unwind to cross a non-POF, but you're right that `mem::forget` is not considered UB. @nikomatsakis and @Amanieu, do you know (/remember) why we consider it UB to skip destructors due to `longjmp` or forced-unwinding but not to skip destructors due to `mem::forget`?

bjorn3:
@BatmanAoD Maybe something like:

> Crates like `rayon` and `take_mut` depend on the fact that function can only return normally or return with a catchable panic to avoid UB. `rayon` requires that all child threads are joined before the parent returns from the `rayon::scope` call and `take_mut` needs to abort on any non normal return to avoid a reference to uninitialized memory. Both would be UB if the forced unwind could cross the `catch_unwind` call. `rayon` also uses a `AbortIfPanic` type to protect against panics in `rayon` itself. If forced-unwinding across destructors was allowed, then `AbortIfPanic` would not help in any way.

BatmanAoD:
@bjorn3 Technically, forced unwinding could execute `drop` functions normally; the problem is that the language mechanisms that use forced unwinding (`pthread_exit` and `longjmp`) don't actually use forced unwinding on most platforms, and other platform implementations simply deallocate the frames rather than unwinding. So if we permitted forced unwinding over non-POFs on platforms where forced unwinding is used, that would mean that e.g. `longjmp` would be well-defined over non-POFs on Windows but not on any other platforms.

rfcs/0000-c-unwind-abi.md
```diff
@@ -126,6 +133,28 @@ may be "sandwiched" between Rust frames, so that Rust `panic`s may safely
...
+types][cpp-POD-definition], which are types that, among other requirements, are
+trivially destructible (their destructors do not cause any observable effects,
+and may be elided as an optimization).
+
+Rust frames that do contain pending destructors or `catch_unwind` calls are
+called non-POFs.
+
+Note that a non-POF may _become_ a POF, for instance if all `Drop` objects are
+moved out of scope, or if its only `catch_unwind` call is in a code path that
+will not be executed.
```
nikomatsakis: I think it'd be worth giving an example here, and I have to admit I still feel sort of hesitant. I wonder if we can call in an LLVM expert to validate this. Let me give a specific example of the situation here:

```rust
fn foo(c: bool) {
    let v = vec![1, 2, 3];
    if c {
        drop(v);
    }
    func_call_that_longjmps(c);
}

/// Only does a longjmp if `c` is true.
extern "C" fn func_call_that_longjmps(c: bool);
```

In this situation, the call to `func_call_that_longjmps` will (on windows, anyway) unwind with a forced exception, despite being marked as "nounwind" and despite the fact that the compile will generate a landing pad (that ultimately has no effect when executing). I believe our claim is that this is not going to be UB.

Actually, I realize as I write this that I haven't really *read* the LLVM definition -- @Amanieu maybe you can provide a citation to the relevant webdocs? I also realize that, from LLVM's POV, the existence of a landing pad is quite likely irrelevant.

So I am starting to be convinced, but I think it would be good for us to include some of this reasoning in the RFC itself.

nikomatsakis:
Regarding why it is UB to unwind a frame:

It is true that `mem::forget` is safe, but that does NOT mean that it is safe not to run destructors. In other words, it's true that unsafe code can **actively cause** destructors not to run, either by using `mem::forget` or by creating a `Rc` cycle. But this doesn't mean that some other code can make that chose on its behalf. Put yet another way, the Rust language offers guarantees (or attempts to) about when destructors will and will not run. It is not a random choice.

In particular, destructors for local variables are guaranteed to run unless the variable is moved or "forgotten". Those destructors in turn will free any data owned by the variable.

(In the case of a Rc cycle, no local variable fully owns the data, and hence it is never freed, but that has more to do with Rc than with Rust the language.)

In practical terms, the reason for this is, yes, that it allows unsafe code like rayon or take-mut to take actions that are guaranteed to execute on unwinding. In other words, this is a long-recognized pattern in unsafe code and to remove it would break big parts of the ecosystem.

rfcs/0000-c-unwind-abi.md
Amanieu: From the [LangRef](https://llvm.org/docs/LangRef.html):
> `nounwind`
>    This function attribute indicates that the function never raises an exception. If the function does raise an exception, its runtime behavior is undefined. However, functions marked nounwind may still trap or generate asynchronous exceptions. Exception handling schemes that are recognized by LLVM to handle asynchronous exceptions, such as SEH, will still provide their implementation defined semantics.

This is a bit vague but basically implies that `longjmp` out of a `nounwind` function has "implementation defined semantics". I have no idea what that is actually supposed to mean though...

nikomatsakis: OK, so I guess there are two considerations:

* On Windows, where we actually *do* generate an exception, what happens. I agree that is vague but suggestive that this is "defined behavior".
* On Linux, where there is no exception, what happens. I believe the idea *here* is that LLVM (unlike Rust) doesn't itself consider it UB to "pop" a frame, even if there are pending cleanups or whatever (e.g., if you called a "C unwind" function which invoked longjmp, that could happen).

BatmanAoD: @nikomatsakis What would be the next step here? Do you still want to find someone with more LLVM expertise (or authority) to confirm that "popping" frames with landing pads is not inherently UB?

nikomatsakis: I'd be curious whether @hanna-kruppe or @RalfJung has some thoughts but at this point I feel fairly comfortable. I think I'd like to amend the text here to include a bit more discussion and detail on this point.

BatmanAoD: @nikomatsakis I've added an example based on your code above, and some explanatory verbiage, including a paragraph about the `nounwind` optimization. Please let me know if what I've written is not what you were looking for!

RalfJung: I'm afraid I know no more than you about LLVM semantics. Might be best to bring this up with the LLVM people? They [have a Discourse](https://llvm.discourse.group/) these days. :D

nikomatsakis: Yes, that seems good.

Issue #30, opened 2020-06-11
follow-up: longjmp annotations and optimization


nikomatsakis:
As discussed in [this week's meeting](https://zulip-archive.rust-lang.org/210922projectffiunwind/76123weeklymeeting.html#200578159), we realized that permitting `longjmp`  out of a `"C"` function will lose optimization potential in a case like this:

```rust
fn foo(x: &mut u32) {
    *x += 1;
    bar();
}
```

At least at present, we should be able to move the `x` down below the call to `bar()`, so long as we do it also on unwinding. However, longjmp would make that observable.

We discussed an idea where people annotate functions that "may longjmp" in some way -- two ideas were `#[longjmp]` and `#[pof]`, though neither is ideal. It would be UB for a function that is to be deallocated unless it carries this annotation. Further:

* The compiler can warn if a `#[longjmp]` function calls another longjmp fn (or a fn that may be longjmp, i.e., by fn pointer) with pending destructors in scope. This carries a risk of false warning since the fn may in fact not longjmp.
* The compiler can warn if a non-longjmp fn calls a longjmp one. This carries a risk of missed warnings since calls by pointer could target a longjmp fn.

We would also suppress reordering optimizations around function calls in longjmp-functions.

Amanieu:
It occurs to me that this optimization might be unsound even without `longjmp`. Consider the case where `x` points to a memory-mapped file and `bar` calls `exit()`. You would expect the write to `x` to be reflected in the file on exit, but that won't happen if the write is moved after the call.

nikomatsakis:
@Amanieu Yes, so I brought up the idea that `&mut` references into shared memory were simply not compatible with this optimization, and I thought the conclusion might be that you should not create `&mut` references into shared memory (you should instead prefer raw pointers or `&Cell`). But memory mapped files are a good example of shared memory in practice, not sure if avoiding `&mut` in such cases is really practical -- maybe it's widespread practice?

BatmanAoD:
I also suggested `#[cancelable]`. `#[cancel-safe]` may be more descriptive. The Linux [pthreads man page](https://linux.die.net/man/7/pthreads) uses the term "async-cancel-safe" for some functions, but I'm not sure what the "async" part means.

We discussed whether an annotation is sufficient and agreed that it probably is. This prevents annotating function pointers as `#[cancelable]`, but that is acceptable (and in any case, annotations for function pointers may eventually be added to the language).

[...]

PR #31, opened 2020-06-12
Make all `longjmp` and `pthread_exit` related behavior TBD

BatmanAoD:
Per #30, the initial `"C unwind"` RFC will not specify when `longjmp` and `pthread_exit` is well-defined.

### DIFF
--- a/rfcs/0000-c-unwind-abi.md
+++ b/rfcs/0000-c-unwind-abi.md
@@ -17,8 +17,8 @@ previously-undefined cases when an unwind operation reaches a Rust function
 boundary with a non-`"Rust"`, non-`"C unwind"` ABI.
 
 As part of this specification, we introduce the term ["Plain Old Frame"
-(POF)][POF-definition]. These are frames that may be safely deallocated with
-`longjmp`.
+(POF)][POF-definition]. These are frames that have no pending destructors and
+can be trivially deallocated.
 
 This RFC does not define the behavior of `catch_unwind` in a Rust frame being
 unwound by a foreign exception. This is something the [project
@@ -78,7 +78,9 @@ how well the current design satisfies these constraints.
   languages) to raise exceptions that will propagate through Rust
   frames "as if" they were Rust panics (i.e., running destrutors or,
   in the case of `unwind=abort`, aborting the program).
-* **Enable error handling with `longjmp`:** As mentioned above, some existing
+* **Enable error handling with `longjmp`:**
+  <!-- XXX TBD -->
+  As mentioned above, some existing
   Rust libraries use `longjmp`. Despite the fact that `longjmp` on Windows is
   [technically a form of unwinding][forced-unwinding], using `longjmp` across
   Rust [POFs][POF-definition] [is safe][longjmp-pr] with the current
@@ -173,7 +175,7 @@ This RFC specifies that, regardless of the platform or the ABI string (`"C"` or
 `"C unwind"`), any platform features that may rely on forced unwinding are:
 
 * _undefined behavior_ if they cross non-[POFs][POF-definition]
-* _defined behavior_ when all unwound frames are POFs
+* _defined behavior_ when all unwound frames are POFs <!-- XXX TBD -->
 
 As an example:
 
@@ -221,8 +223,9 @@ This RFC retains most of that undefined behavior, with two exceptions:
 * With the `panic=unwind` runtime, `panic!` will cause an `abort` if it would
   otherwise "escape" from a function defined with `extern "C"`.
 * Forced unwinding is safe with `extern "C"` as long as only
-* [POFs][POF-definition] are unwound. This is to keep behavior of
+  [POFs][POF-definition] are unwound. This is to keep behavior of
   `pthread_exit` and `longjmp` consistent across platforms.
+  <!-- XXX emphasize this is not TBD! -->
 
 ## Interaction with `panic=abort`
 
@@ -255,6 +258,7 @@ behavior. `"C"`-like ABIs are `"C"` itself but also related ABIs such as
 | `panic=abort`  | `"C unwind"` | `panic!` aborts                       | abort                   |
 | `panic=abort`  | `"C"`-like   | `panic!` aborts (no unwinding occurs) | UB                      |
 
+<!-- XXX TBD
 Regardless of the panic runtime, ABI, or platform, the interaction of Rust
 frames with C functions that deallocate frames (i.e. functions that may use
 forced unwinding on specific platforms) is specified as follows:
@@ -262,6 +266,7 @@ forced unwinding on specific platforms) is specified as follows:
 * **When deallocating Rust [POFs][POF-definition]:** frames are safely
   deallocated; no undefined behavior
 * **When deallocating Rust non-POFs:** undefined behavior
+-->
 
 No subtype relationship is defined between functions or function pointers using
 different ABIs. This RFC also does not define coercions between `"C"` and
@@ -389,11 +394,13 @@ future work.
 
 ### Enable error handling with `longjmp`
 
+<!-- XXX TBD -->
 This constraint is met: `longjmp` is treated the same across all platforms, and
 is safe as long as only [POFs][POF-definition] are deallocated.
 
 ### Do not change the ABI of functions in the `libc` crate
 
+<!-- XXX TBD -->
 This constraint is met: `libc` functions will continue to use the `"C"` ABI.
 `pthread_exit` will be treated the same across all platforms, and will be safe
 as long as only [POFs][POF-definition] are deallocated. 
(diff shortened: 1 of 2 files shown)


rfcs/0000-c-unwind-abi.md
```diff
@@ -485,9 +434,15 @@ provide a well-defined behavior for this case, which will probably be either to
 let the exception pass through uncaught or to catch some or all foreign
 exceptions.
 
+We would also like to specify conditions under whch `longjmp` and
```
bjorn3:
```suggestion
We would also like to specify conditions under which `longjmp` and
```

rfcs/0000-c-unwind-abi.md
```diff
@@ -57,11 +51,11 @@ several requirements for any cross-language unwinding design.
...
 how well the current design satisfies these constraints.
 
-* **Changing from panic=unwind to panic=abort cannot cause UB:** We
-  wish to ensure that choosing `panic=abort` doesn't ever create
-  undefined behavior (relate to `panic=unwind`), even if one is
-  relying on a library that triggers a panic or a foreign exception.
-* **Optimization with panic=abort:** when using `-Cpanic=abort`, we
+* **Changing from `panic=unwind` to `panic=abort` cannot cause undefined
+  behavior:** We wish to ensure that choosing `panic=abort` doesn't ever create
+  undefined behavior (relate to `panic=unwind`), even if one is relying on a
```
Ixrec: It's not immediately clear to me what this parenthetical means. Maybe "changing from `panic=unwind` to `panic=abort` never creates undefined behavior", if that's the intent?

nikomatsakis: Agreed that phrasing is clearer

rfcs/0000-c-unwind-abi.md
```diff
@@ -78,19 +72,20 @@ how well the current design satisfies these constraints.
...
-  Rust libraries use `longjmp`. Despite the fact that `longjmp` on Windows is
-  [technically a form of unwinding][forced-unwinding], using `longjmp` across
-  Rust [POFs][POF-definition] [is safe][longjmp-pr] with the current
-  implementation of `rustc`, and we want to specify that this will remain safe.
+* **Enable error handling with `longjmp`:**
+  As mentioned above, some existing Rust libraries use `longjmp`. Despite the
+  fact that `longjmp` on Windows is [technically a form of
+  unwinding][forced-unwinding], using `longjmp` across Rust
+  [POFs][POF-definition] [is safe][longjmp-pr] with the current implementation
+  of `rustc`, and we would like this to remain safe.
```
nikomatsakis: I would say something like

"using `longjmp` across Rust POFs is something that existing libraries rely upon today to interoperate with Ruby, Lua, and other C APIs based on longjmp. Althogh this RFC doesn't fully enable this use case, it is meant to be 'forwards compatible' with a future RFC that will define the correct way to interoperate with longjmp-based APIs. (That ffi-unwind working group intends to pursue this RFC.)"

rfcs/0000-c-unwind-abi.md
```diff
@@ -216,29 +180,20 @@ boundary, either from a `panic!` "escaping" from a Rust function defined with
...
-There are some types of unwinding that are not guaranteed to cause the program
-to abort with `panic=abort`, though:
-
-* Forced unwinding: Rust provides no mechanism to catch this type of unwinding.
-  This is safe with either the `"C"` ABI or the new `"C unwind"` ABI, as long
-  as only [POFs][POF-definition] are unwound.
-* Unwinding from another language into Rust if the entrypoint to that language
-  is declared with `extern "C"` (contrary to the guidelines above): this is
-  always undefined behavior.
+However, unwinding from another language into Rust through an FFI entrypoint
```
nikomatsakis: Maybe something like:

> With the exception of the above case, however, unwinding from another language into Rust through an FFI entrypoint declared with `extern "C"` is always undefined behavior, and is not guaranteed to cause the program to abort under `panic=abort`.

An alternative of course would be to say that the old text is correct, it's just that longjmp and pthread-exit **may not unwind**, so using those on platforms where they have different behavior is potentially UB.

rfcs/0000-c-unwind-abi.md
BatmanAoD: Oops, I think I copied this bullet from a conversation somewhere and didn't read it closely. Glad that adding the backticks happened to draw attention to it!

BatmanAoD:
```suggestion
  behavior:** We wish to ensure that changing from `panic=unwind` to `panic=abort` never creates undefined behavior (relate to `panic=unwind`), even if one is relying on a
```

rfcs/0000-c-unwind-abi.md
```diff
@@ -255,13 +210,10 @@ behavior. `"C"`-like ABIs are `"C"` itself but also related ABIs such as
...
-frames with C functions that deallocate frames (i.e. functions that may use
-forced unwinding on specific platforms) is specified as follows:
-
-* **When deallocating Rust [POFs][POF-definition]:** frames are safely
-  deallocated; no undefined behavior
-* **When deallocating Rust non-POFs:** undefined behavior
+The interaction of Rust frames with C functions that deallocate frames (i.e.
+functions that may use forced unwinding on specific platforms) is not specified
+in this RFC. However, we do specify that deallocating
+non-[POFs][POF-definition]s in this way is undefined behavior.
```
nikomatsakis: I sort of like the old setup, which I found somewhat clearer. Can we just change the "When deallocating Rust POFs" to say "this RFC does not fully specify this behavior. For now, it must be considered undefined, but the project-ffi-unwind group expects to create some path that enables POF frames to be safely and correctly deallocated via longjmp. However, additional annotations may be required."

rfcs/0000-longjmp-pof-annotation.md
```diff
@@ -0,0 +1,85 @@
...
+#[cancelable]
+fn foo<D: Drop>(c: bool, d: D) {
+  if c {
+    drop(d);
+  }
+  longjmp_if_true(c);
+}
+
+/// Calls `longjmp` if `c` is true; otherwise returns normally.
+extern "C" fn longjmp_if_true(c: bool);
```
nikomatsakis, line 40: I think I would expect this function to have a `#[cancelable]` attribute.

nikomatsakis: That does tend to make the name "cancelable" feel less good, though.

rfcs/0000-longjmp-pof-annotation.md
```diff
@@ -0,0 +1,85 @@
...
+form of unwinding: the `foo` frame is simply discarded. On Windows, `longjmp`
+is implemented as a forced unwind, which is permitted to traverse `nounwind`
+frames. Since `foo` contains a `Drop` type the forced unwind will include a
+call to the frame's cleanup logic, but that logic will not produce any
+observable effect; in particular, `D::drop()` will not be called again. The
+observable behavior should therefore be the same on all platforms.
+
+Conversely, if, due to a bug, `longjmp` were called unconditionally, then this
+code would have undefined behavior on all platforms when `c` is false, because
+`foo` would not be a POF.
```
nikomatsakis, line 58: I would expect that the compiler would warn in this scenario, also

rfcs/0000-longjmp-pof-annotation.md
BatmanAoD: Yes, I think it does make sense to require `extern` functions to be annotated as well. Fortunately, I think we have plenty of time to bikeshed a name for the annotation!

rfcs/0000-longjmp-pof-annotation.md
BatmanAoD: Maybe I'm misunderstanding, but how would the compiler provide a warning for this? The `extern` function is in C, so the compiler can't determine under what conditions it will `longjmp`. Or do you mean it would warn in any scenario where a frame is not guaranteed to be a POF at the time it calls a "cancelable" function?

rfcs/0000-c-unwind-abi.md
BatmanAoD: I think alluding to new annotations in this RFC would be premature, but I'll add something along those lines.

nikomatsakis: The main thing is "you may have to change the source in some way"

rfcs/0000-longjmp-pof-annotation.md
nikomatsakis: > do you mean it would warn in any scenario where a frame is not guaranteed to be a POF at the time it calls a "cancelable" function?

This is what I meant. I think it's going to be quite rare that you have a function that *sometimes cancels* but you happen to know that *on this particular call it will not* and therefore you want to take advantage of that by having a pending destructor. In such a case, adding an `#[allow]` annotation seems reasonable to me.

nikomatsakis (approved):

rfcs/0000-c-unwind-abi.md
nikomatsakis: The text you added seems fine.

rfcs/0000-longjmp-pof-annotation.md
BatmanAoD: Adding a comment to that effect

